Issue #469, opened 2014-12-23
Tags appear to be added twice


kyllikekimmel:
When I add a tag to my time entry and continue the same entry later, the tag that I've applied earlier appears to have been added twice. 

Android 7.0.19

monday8am:
out of date. Tag component and common code totally replaced in Android.

PR #470, opened 2014-12-23
Fix duplicate tag save tasks.

roosmaa:
This what was fixed for Ross in 7469d57.

Might be what's causing #469, need to investigate this possibility.


monday8am:
:+1:

taavirehemagi:
:+1:  from me

PR #471, opened 2015-01-02
Added tap-zoom to bars in Joey's reports. Closes #425

monday8am:
Added methods to make zoom tapping on bars. Fixes #425

PR #472, opened 2015-01-03
Tap-zoom bar on Joey's reports using ViewPropertyAnimatior. Closes #425

### LINKED ISSUE #425: Tap-to-zoom for barchart month view
Bar tap @ month view zooms three closest rows to full(normal) height and makes durations visible.
Epic #178

monday8am:
New method of animation using ViewPropertyAnimator. Compared with previous solution (https://github.com/toggl/mobile/pull/471) , this new has better performance.

requestlayout

translatex

the charts were calculated using the method described here: http://www.curious-creature.com/2012/12/01/android-performance-case-study/comment-page-1/

I used the existing method to calculate data (top positions and height) and changed only the animation methods, so, floating point is avoided whenever possible.

See #425

### DIFF
--- a/Joey/UI/Views/BarChart.cs
+++ b/Joey/UI/Views/BarChart.cs
@@ -28,12 +28,14 @@ namespace Toggl.Joey.UI.Views
         private int rightPadding;
         private int yAxisSpacing;
         private int rowMargin;
+        private int selectedRowIndex;
         private int barZeroSize;
         private int barLabelSpacing;
         private BackgroundView backgroundView;
         private View loadingOverlayView;
         private View emptyOverlayView;
         private Animator currentAnimation;
+        private PointF tapInitialPos;
         private bool isZooming;
 
         public BarChart (Context context, IAttributeSet attrs) : base (context, attrs)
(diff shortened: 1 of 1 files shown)


roosmaa:
Looks awesome! Just 1 request, could you tweak `GetSelectedBarIndex` to allow for some error margin? 

Basically currently when I happen to tap between two bars nothing happens. It would be nice if this didn't exist, so `GetSelectedBarIndex ` should find the closest bar to the touch point.

Also, when I touch on the bottom axis for instance, `GetSelectedBarIndex` should return -1, so we should have some check that if the distance from the closest bar is too long, we treat it as no bars found.

monday8am:
nice! I´ll fix it :hammer:

monday8am:
I added few lines of codes :) .. but I think is not bad. With this addition there is not space between bars... so, the char will always react to touches in any place. If touch is clearly inside a zoomed bar, the chart will be zoomed out.

monday8am:
yeapp, totally agree.

taavirehemagi:
Looks good to me!

roosmaa:
:+1:

PR #486, opened 2015-01-13
BaseTracker Test

taavirehemagi:


roosmaa:
I think there are options to specify that the test should expect an exception (an extra Attribute for the test method) and there should also be an Assert method which expects the delegate to raise an exception. Maybe that would make the tests a bit cleaner?

taavirehemagi:
Made some changes to some parts, maybe it's a bit more clear now. What do you think?

roosmaa:
Well.. tests are failing now :p

taavirehemagi:
The failing was fixed. What's the status on this one?  :)

monday8am:
BaseTracker have methods to send info related to the data passed. The implementation of how to send this info is not concern to this test.  So,

- Create a based tracker: ok.
- Extend the methods: ok ( in my opinion, you don´t need to raise exceptions)
- Check if logic is working: for each case, check that if you pass data A, the correct string is sent. 
- Check if correct exception is launched when the data is not correct: ok ( but not in the most direct way)

I think you should need to add every case, and try to keep the test code as simple as possible. @roosmaa , could you let us know your op. in this case?

roosmaa:
It's such a stupid class, I'm happy with 100% coverage of BaseTracker (all of the lines of codes get executed at least once). But it's your call @monday8am , if you want @taavirehemagi to spend more time on it or not.

If @monday8am doesn't object, I'm giving it :+1:

monday8am:
@roosmaa I learn with discussions about test, more if you´re in. If I learn with a simple class, I´ll do better with complex, probably Taavi as well ;) :) , here is my kudo then :+1:

Issue #485, opened 2015-01-13
Project duration displays wrong on reports listview


taavirehemagi:
In case of classic duration format, Hours instead of TotalHours are displayed. 
![screen shot 2015-01-13 at 14 18 03](https://cloud.githubusercontent.com/assets/2154171/5720608/2a1d2e02-9b2f-11e4-820a-bdfc9f16ca42.png)

PR #492, opened 2015-01-23
Correct format of TimeEntries on Reports. Closes #488, #485

### LINKED ISSUE #488: Errors on Bartchar due to misreading of Report data
Seems to be theres is a lack of understanding between app and Reports API. The problem is present on data that feed the Barchart. 
Could be simulated with test account and year zoom level ( the piechart shows data but the barchart not).

monday8am:
- Correct format.
- Now Barchart labels are formatted with the format selected by the user ( with minor modifications). The final result is equal to web barchart.

Fixes #488, #485

### DIFF
--- a/Phoebe/Data/Models/TimeEntryModel.cs
+++ b/Phoebe/Data/Models/TimeEntryModel.cs
@@ -229,7 +229,7 @@ namespace Toggl.Phoebe.Data.Models
         public string GetFormattedDuration ()
         {
             TimeSpan duration = GetDuration (Data, Time.UtcNow);
-            string formattedString = duration.ToString (@"h\:mm\:ss");
+            string formattedString = duration.ToString (@"hh\:mm\:ss");
             var user = ServiceContainer.Resolve<AuthManager> ().User;
 
             if (user == null) {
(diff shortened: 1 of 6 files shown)


Phoebe/Data/Models/TimeEntryModel.cs
```diff
@@ -229,7 +229,7 @@ public bool IsBillable
         public string GetFormattedDuration ()
         {
             TimeSpan duration = GetDuration (Data, Time.UtcNow);
-            string formattedString = duration.ToString (@"h\:mm\:ss");
+            string formattedString = duration.ToString (@"hh\:mm\:ss");
```
roosmaa: Why this change? Do we want to display duration in "08:00:00" format? Will UI survive the extra 0?

monday8am: I searched and your previous uses of this format include the two "hh". 

https://github.com/toggl/mobile/blob/314cbd59544b372c126d544c75487f22c72ab6a7/Ross/ViewControllers/TimerNavigationController.cs
https://github.com/toggl/mobile/blob/e6ab7a7880bd4284ef0c7e120ea7f74a3d45ef28/Joey/UI/Adapters/LogTimeEntriesAdapter.cs

The UI is prepared to accept two digits in this format ( 12 for example).

roosmaa:
Looks OK. Didn't try to reproduce the bugs to see if it actually fixes them - @taavirehemagi can you do that? :angel:

taavirehemagi:
OK from me. Fixes the bugs. :+1:

Issue #498, opened 2015-02-03
New logout message for widget


monday8am:
Add new message or notification if the user is not authenticated. @pe0ny I'll need your help for that :)

![screen shot 2015-02-02 at 20 03 53](https://cloud.githubusercontent.com/assets/1842773/6018831/2d218398-ab9a-11e4-9616-bf505b15315f.png)

Epic: #265

monday8am:
Solution defined  by @pe0ny 

![toggl-ios-widget-login](https://cloud.githubusercontent.com/assets/1842773/6104706/2ef75f52-b050-11e4-80d9-dfcc86b889ca.png)

monday8am:
Closed by #501

Issue #480, opened 2015-01-06
Add correct design to existing widget.


monday8am:
Add correct look & feel to existing widget.

Epic #265

monday8am:
Closed by #501

Issue #479, opened 2015-01-06
Implement all required functionality on iOS widget


monday8am:
Implement all required functionality on widget:

- Show last 5 projects used by user and the project stated.
- Sync project state with server.
- On tap, open Toggl app and start / stop this projects. 

Epic: #265

monday8am:
Closed by #501

Issue #265, opened 2014-07-19
iOS8 notification area widget


roosmaa:
Explore the possibility of implementing a running time entry display widget for iOS8 users.

monday8am:
Closed by #501

PR #501, opened 2015-02-04
Today widget implementation ( Emma) Closes #484 #480 #479 #498

### LINKED ISSUE #484: Communicate Toggl app(Ross) with Today widget(Emma)
Find the more optimised way to transfer data between both apps. This data should be updated every second.

Epic: #265

monday8am:
Features:
- Show last 3 - 4 projects.
- Able to Start - Stop - Continue this projects.

Fixes #484 #480 #479 #498  
Epic: #265


Emma/Views/StartStopBtn.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            var posX = (rect.Width - radius) / 2;
+            var posY = (rect.Height - radius) / 2;
+
+            using (var context = UIGraphics.GetCurrentContext ()) {
+
+                UIBezierPath ovalPath;
+                UIBezierPath bezierPath;
+
+                if ( isActive) {
```
masterrr: You should fix this weird “spaces after brackets” format across the pr :)

roosmaa:
In logged out state I can select the text?!

ios simulator screen shot 09 feb 2015 13 30 32


Though it's probably not a problem since the logout state will be a slightly different UI anyway?

But it is taking up a lot of space ... would it be possible to resize it to occupy less space when logged out?

Also, updating is slighly laggy - I open notification area and then it takes ~1s for the widget to update to latest state. Otherwise works nice!

.gitignore
```diff
@@ -15,3 +15,4 @@ local.properties
 *.suo
 packages
 Resource.designer.cs
+DummyContainer
```
roosmaa: What's that being ignored?

monday8am:
All widgets suffers the same delay, in my opinion. 

The logout UI will be fixed with #498

Emma/Views/StartStopBtn.cs
monday8am: :) I´ll fix it in my next PR.

@masterrr  Apart of this tiny detail, did you checked / tested the code?

masterrr:
Master changes should be back to this branch

monday8am:
@masterrr Yes, I should rebase it with master or fix conflicts before merge. Is that what you're saying?

Emma/Views/WidgetCell.cs
```diff
@@ -0,0 +1,243 @@
...
+                },
+            };
+
+            textContentView.Layer.Mask = maskLayer;
+
+        }
+
+        public override void UpdateConstraints ()
+        {
+            if ( ContentView.Constraints.Length > 0) {
```
masterrr: I don't think it's the right approach: few lines below we're operating with constraints on a data state basis. If the data swaps its state after the constraints are set once, they won't be updated.

---
Even if we go with this piece of code, we need to call base.UpdateConstraints() here, before the return statement.

masterrr:
@monday8am yes, everyone who wills to test it gets the failed merge :)

Emma/Views/WidgetCell.cs
```diff
@@ -0,0 +1,243 @@
...
+                    descriptionLabel.WithSameLeft ( projectLabel),
+                    descriptionLabel.Below ( projectLabel, 0f),
+                    descriptionLabel.AtBottomOf (textContentView, 10f),
+                    null
+                );
+            }
+
+            base.UpdateConstraints ();
+
+            LayoutIfNeeded();
```
masterrr: Don't think it's needed here

Ross/Data/WidgetUpdateService.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            Guid entryId;
+            Guid.TryParse (UserDefaults.StringForKey (StartedEntryKey), out entryId);
+            return entryId;
+        }
+
+        public void UpdateWidgetContent()
+        {
+            var controller = NCWidgetController.GetWidgetController ();
+            controller.SetHasContent (true, "com.toggl.timer.today");
```
masterrr: Maybe fetch the prefix from dynamic NSBundle.MainBundle.BundleIdentifier?

Ross/Data/WidgetUpdateService.cs
```diff
@@ -0,0 +1,104 @@
...
+        public static string ContinueEntryUrlPrefix = "continue";
+
+        private NSUserDefaults nsUserDefaults;
+        private UIViewController rootController;
+
+        public NSUserDefaults UserDefaults
+        {
+            get {
+                if ( nsUserDefaults == null) {
+                    nsUserDefaults = new NSUserDefaults ("group.com.toggl.timer", NSUserDefaultsType.SuiteName);
```
masterrr: NSBundle.MainBundle.BundleIdentifier too?

monday8am: I don't want to use the MainBundle. To share data between widget and app, I should use a shared container created with the name of the App Group (group.com.toggl.timer) that contains both the widget and the app.

Ross/Data/WidgetUpdateService.cs
monday8am: In this line, I'm saying to widget (com.toggl.timer.today)  that new content is available.

Emma/Views/WidgetCell.cs
monday8am: I call LayoutIfNeeded() to apply the mask correctly. The mask should be applied when subviews have been aligned and resized by constrains. Let me know if you have a better approach.

Emma/Views/WidgetCell.cs
monday8am: got it. And which is your suggestion? Remove and add Constrains every time? About second sentence, agreed.

taavirehemagi:
Works good for me!

Ross/Data/WidgetUpdateService.cs
masterrr: I thought concatenation (group.+%BundleIdentifier%)

Emma/Views/WidgetCell.cs
masterrr: > Remove and add Constrains every time? 
That would work, but the performance will be poor

I'm thinking about deleting/adding/tracking "data.isEmpty"-constraints (those which are being added in if/else clause)

Ross/Data/WidgetUpdateService.cs
masterrr: ok

Ross/Ross.csproj
```diff
@@ -67,6 +67,7 @@
     <MtouchUseSGen>true</MtouchUseSGen>
     <MtouchUseRefCounting>true</MtouchUseRefCounting>
     <MtouchLink>None</MtouchLink>
+    <CodesignProvision>11507202-61c0-4a57-99c6-c5f5a9f8bd49</CodesignProvision>
```
masterrr: Delete your provision :)

monday8am: :+1:

masterrr:
Widget is saying “Log to track time” though I'm already logged in (app is live underneath, on the screenshot). I noticed blurred background is not taking all the widget space so there is a little gap without blurry effect in the bottom, is it for the purpose or need to be fixed? 
screen shot 2015-02-16 at 9 55 31 pm

masterrr:
After using the app for a while the “login to track time” section disappeared and now I've just got an empty section. (I have entries in the app, timer is not running)

screen shot 2015-02-16 at 10 00 10 pm

monday8am:
@masterrr this errors were due a problem with the entitlements file in your project.

monday8am:
I have added support for background / closed app:
- If the app go to background, is not possible to update the time counter. A simple solution is to hide it until the app becomes active again.
- If the app is closed, the widget only shows a message to open the app again ( similar to the login message).

For future releases, a good improvement could be add background features to the app and add an internal counter to the widget.

Emma/Views/WidgetCell.cs
```diff
@@ -0,0 +1,244 @@
...
+            base.LayoutSubviews ();
+        }
+
+        private UIColor UIColorFromHex ( string hexValue, float alpha = 1f)
+        {
+            hexValue = hexValue.TrimStart ('#');
+
+            int rgb;
+            if (!Int32.TryParse (hexValue, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out rgb)) {
+                throw new ArgumentException ("Invalid hex string.", "hexValue");
```
masterrr: maybe we should go easy on the wrong value and fallback to some neutral default project color?

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+            openAppView.Add (textView = new UILabel {
+                TranslatesAutoresizingMaskIntoConstraints = false,
+                Font = UIFont.FromName ("Helvetica", 13f),
+                Text = isAppActive ? "NoLoggedUser".Tr() : "NoActiveApp".Tr(),
+                TextColor = UIColor.White,
+                BackgroundColor = UIColor.Clear,
+            });
+
+            openAppView.Add (openAppBtn = new StartStopBtn {
+                TranslatesAutoresizingMaskIntoConstraints = false,
```
masterrr: Consider using openAppView.TranslatesAutoresizingMaskIntoConstraints(); after adding the subviews instead of setting false on every view manually

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+        {
+            base.ViewDidLoad ();
+
+            openAppBtn.TouchUpInside += (sender, e) => {
+                if (WidgetDisabled) {
+                    return;
+                }
+                WidgetDisabled = true;
+                openAppBtn.IsActive = false;
+                UIApplication.SharedApplication.OpenUrl (new NSUrl ("com.toggl.timer://" + TodayUrlPrefix + "/"));
```
masterrr: Consider BundleIdentifier instead of the hard-coded value

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+            };
+
+            source.OnContinue += (sender, e) => {
+                if (WidgetDisabled) {
+                    return;
+                }
+                WidgetDisabled = true;
+                var id = string.IsNullOrEmpty (source.SelectedCellId) ? new Guid().ToString() : source.SelectedCellId;
+                UserDefaults.SetString (id, StartedEntryKey);
+                UIApplication.SharedApplication.OpenUrl (new NSUrl ("com.toggl.timer://" + TodayUrlPrefix + "/" + ContinueEntryUrlPrefix));
```
masterrr: BundleIdentifier

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+            }
+
+            var source = new TableDataSource (entries);
+
+            source.OnStartStop += (sender, e) => {
+                if (WidgetDisabled) {
+                    return;
+                }
+                WidgetDisabled = true;
+                UIApplication.SharedApplication.OpenUrl (new NSUrl ("com.toggl.timer://" + TodayUrlPrefix + "/" + StartEntryUrlPrefix));
```
masterrr: BundleIdentifier

Phoebe/WidgetSyncManager.cs
```diff
@@ -0,0 +1,253 @@
...
+                var store = ServiceContainer.Resolve<IDataStore> ();
+                var userId = ServiceContainer.Resolve<AuthManager> ().GetUserId ();
+
+                var baseQuery = store.Table<TimeEntryData> ()
+                                .OrderBy (r => r.StartTime, false)
+                                .Where (r => r.DeletedAt == null
+                                        && r.UserId == userId
+                                        && r.State != TimeEntryState.New
+                                        && r.StartTime >= queryStartDate)
+                                .Take ( 5);
```
masterrr: Should take 3 or 4, I guess

Phoebe/WidgetSyncManager.cs
```diff
@@ -0,0 +1,253 @@
...
+        private async void OnSync (Message msg)
+        {
+            if ( isLoading) {
+                return;
+            }
+
+            isLoading = true;
+
+            try {
+                var queryStartDate = Time.UtcNow - TimeSpan.FromDays (9);
```
masterrr: Expand days interval? Why 9?

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+                }
+                WidgetDisabled = true;
+                var id = string.IsNullOrEmpty (source.SelectedCellId) ? new Guid().ToString() : source.SelectedCellId;
+                UserDefaults.SetString (id, StartedEntryKey);
+                UIApplication.SharedApplication.OpenUrl (new NSUrl ("com.toggl.timer://" + TodayUrlPrefix + "/" + ContinueEntryUrlPrefix));
+            };
+
+            tableView.RegisterClassForCellReuse (typeof (WidgetCell), WidgetCell.WidgetProjectCellId);
+            tableView.Source = source;
+            tableView.Delegate = new TableViewDelegate ();
```
masterrr: What do you thinking about doing all the instantiating in the LoadView() method, making source an i-var?

Emma/WidgetViewController.cs
```diff
@@ -0,0 +1,354 @@
...
+                    var emptyEntry = new WidgetEntryData { IsEmpty = true };
+                    entries.Insert (0, emptyEntry);
+                    entries.RemoveAt (entries.Count - 1);
+                }
+
+            } else {
+                entries.Add (new WidgetEntryData { IsEmpty = true });
+            }
+
+            var source = new TableDataSource (entries);
```
masterrr: Make a source an i-var, instantiate it in the loadView() method and left something like source.UpdateEntries(entries) here? What's your opinon?

Phoebe/WidgetSyncManager.cs
monday8am: "Take 5" sounds better, you don't like jazz? ;)
Fixed to 4.

Phoebe/WidgetSyncManager.cs
monday8am: To maintain correspondency with data taken on AllTimeEntriesView and showed on the log screens:
https://github.com/toggl/mobile/blob/master/Phoebe/Data/Views/AllTimeEntriesView.cs

PR #582, opened 2015-03-31
Swipe delete implementation.

### LINKED ISSUE #540: Display delete button on list item swipe.
Epic #443

taavirehemagi:
Closes #540

### DIFF
--- a/Joey/Joey.csproj
+++ b/Joey/Joey.csproj
@@ -156,6 +156,8 @@
     <Compile Include="UI\Utils\RecycledBindableViewHolder.cs" />
     <Compile Include="UI\Adapters\LogTimeEntriesAdapter.cs" />
     <Compile Include="UI\Fragments\LogTimeEntriesListFragment.cs" />
+    <Compile Include="UI\Views\ListItemSwipeable.cs" />
+    <Compile Include="UI\Utils\SwipeDeleteTouchListener.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="Resources\AboutResources.txt" />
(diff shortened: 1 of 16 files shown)


monday8am:
- the general solution is ok in the first test ( a listener, a base item, scroll events)
- the animation will need a review but is ok. 
- needs to be rebase with last changes in android-redesign branch.
- but, the event OnDismiss is not connected with the code! (is never called) :) ..

taavirehemagi:
Rebased from android-redesign. I planned the OnDismiss to be used once the RecyclerView for home screen is ready to have some place to attach it to. If I attach it to current adapter, it will be changed anyway, no?

monday8am:
The touch listener will be almost the same. You only need to detect the end of the movement here:
https://github.com/toggl/mobile/blob/taavirehemagi-redesign-swipedismiss/Joey/UI/Utils/SwipeDeleteTouchListener.cs#L87

...and dispatch the OnDismiss event. The animation code is encapsulated in the ListItemSwipeable, try to find a way to notify from the item to the touch listener or move the animation code to the touch listener (like the original code :D)

taavirehemagi:
So you don't mean that I should make the button actually delete data, but just to have the OnDismiss dispatching the end of event? Of course, I will add the OnDismiss to dispatch at end of the animation, (I somehow missed that before). Do you need me to add the actual data deletion as well?

monday8am:
This is the chain of actions:
- Received OnDismiss.
- Get correct TimeEntryModel from RecyclerView using the position received and delete it from list (not from database).
- Show the Undo bar for 3 seconds.
- In case of press Undo, insert the element in same position.
- Hide the Undo bar.
- Try to Delete the TimeEntry from database.

You're in charge of the Undo bar but but just to show / hide it. I'll take care of the rest of actions. The last action is ready from previous code.

taavirehemagi:
got it

taavirehemagi:
Added undo support. Check it out

monday8am:
superseded by #608!

Issue #623, opened 2015-04-28
Group TE: tags view binding, bulk update


masterrr:

masterrr:
Related: https://github.com/toggl/mobile/pull/686

masterrr:
Closed by https://github.com/toggl/mobile/pull/686

Issue #656, opened 2015-05-04
Hight cpu use on Grouped TE Activity load


monday8am:
https://github.com/toggl/mobile/blob/android-redesign/Joey/UI/Activities/EditTimeEntryActivity.cs#L62

This code should be redefined. Maybe moving it to Phoebe and not using loaded models.

monday8am:
fixed at #665

PR #667, opened 2015-05-07
TE Edit (single, group) enhancements

masterrr:
toggl_group_te

1. New functionality and refactoring in Tags Dialog, Project List, Create Tag Dialog: Group (bulk) and Single update (using ITimeEntryModel);
2. Refactor EditGroupedTimeEntryFragment;
3. !Bridge Object to pass objects between activities! (NavBridge), used by ProjectListActivity and BaseEditTimeEntryFragment;
4. Updated ITimeEntryModel and corresponding TimeEntryModel and TimeEntryGroup with an “Apply” method to execute on entities

Closing https://github.com/toggl/mobile/issues/623


Joey/UI/Fragments/BaseEditTimeEntryFragment.cs
```diff
@@ -393,7 +393,12 @@ private void OnProjectEditTextClick (object sender, EventArgs e)
             }
 
             var intent = new Intent (Activity, typeof (ProjectListActivity));
-            intent.PutExtra (ProjectListActivity.ExtraTimeEntryId, TimeEntry.Id.ToString ());
+
+            NavBridge.FinishedNav = projectListActivity => {
+                ProjectListActivity activity = (ProjectListActivity)projectListActivity;
+                activity.Model = model;
+            };
+
```
monday8am: if you want to detect if an activity has finished, user this: http://developer.android.com/training/basics/intents/result.html

Phoebe/Data/Utils/TimeEntryGroup.cs
```diff
@@ -172,26 +175,30 @@ public Task StopAsync ()
...
-
-            TimeEntryModel mModel;
-            for (int i = 0; i < dataObjects.Count; i++) {
-                mModel = new TimeEntryModel (dataObjects [i]);
-                await mModel.DeleteAsync ();
-            }
+            await TTask.Run (() => Parallel.ForEach (dataObjects, obj => {
+                var m = new TimeEntryModel (obj);
+                m.DeleteAsync();
+            }));
```
monday8am: is more fast this way? in case of yes, :+1: for me. I was thinking to reproduce all code of TimeEntryModel in this place.

Joey/UI/Fragments/ChooseTimeEntryTagsDialogFragment.cs
```diff
@@ -43,22 +39,11 @@ public ChooseTimeEntryTagsDialogFragment (IntPtr jref, Android.Runtime.JniHandle
...
-        private Guid TimeEntryId
-        {
-            get {
-                var id = Guid.Empty;
-                if (Arguments != null) {
-                    Guid.TryParse (Arguments.GetString (TimeEntryIdArgument), out id);
-                }
-                return id;
-            }
-        }
```
monday8am: This is a kind of persistence to get the Id when the fragment is re-created from a parameterless constructor. Why you remove it?

http://stackoverflow.com/questions/10450348/do-fragments-really-need-an-empty-constructor

Phoebe/Data/Utils/TimeEntryGroup.cs
masterrr: more reliable, the loop sometimes doesn't update all entries for me

Joey/UI/Fragments/BaseEditTimeEntryFragment.cs
masterrr: It doesn't detect if an activity has finished, but detects when it (projectListActivity) started and updates the public model property

Phoebe/Data/Utils/TimeEntryGroup.cs
masterrr: using await Task.Run() layer to conform with the interface

Joey/UI/Fragments/ChooseTimeEntryTagsDialogFragment.cs
masterrr: Good point. Just tested with “Don't keep activities” developer option on, the fragment is non-functional. Fixing

Joey/UI/Fragments/BaseEditTimeEntryFragment.cs
monday8am: I understand now. 
This way to set the Model to the activity is not functional. The activity must be self-sufficient to fill its content, just using info passed on intent. In this way, the system could kill the activity safely if more memory is needed and restarts it from the intent later.

If you want to pass complex objects read this:
http://stackoverflow.com/questions/4878159/android-whats-the-best-way-to-share-data-between-activities 

And more important, every part of the code is improvable but each part was defined and tested consciously by a programmer before you arrived. Try to find very good reasons every time you want to delete something.

masterrr: Agreed, applied iOS thinking here. Just completed self-sufficient restoring args part for tags fragment. Will port the approach to the project list activity tomorrow.

P.S. Though this approach doesn't fail with “don't keep activities” on, activity works and loads after switching to the app from the recent screen well, will do more testing tomorrow

monday8am: > Though this approach doesn't fail with “don't keep activities” on, activity works and loads after switching to the app from the recent screen well.
Android decides about to kill the activity or not depending of memory needs. Maybe your test works now but fails later or for other users/phones/conditions. 

> will do more testing tomorrow
?? Your tasks this two weeks are to close issues, not to test how activities works. Close your issues and let tests and learning for your free time.

masterrr: This is a PR for the issue https://github.com/toggl/mobile/issues/623

> Android decides about to kill the activity or not depending of memory needs. Maybe your test works now but fails later or for other users/phones/conditions.
That's right

> Not to test how activities lifecycle
Lack of it led me to break the self-sufficient fragment property, nevertheless I'm doing pure learning in my spare time

masterrr:
@monday8am all the bugs (especially activities/fragments restoration from the intent and args) are fixed now

taavirehemagi:
OK from me

Phoebe/Data/Utils/TimeEntryFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+                return new TimeEntryModel (ids.First ());
+            } else {
+                var grp = new TimeEntryGroup ();
+                await grp.BuildFromGuids (ids);
+                return grp;
+            }
+        }
+    }
+}
+
```
monday8am: @masterrr: 
- One static method don't justify the creation of a new class. Less classes, less code, more easy to understand.

- Avoid the creation of Models. If you create a new TimeEntryModel, you're filling a whole object with client, tasks, projects and you'll receive events about changes, etc. Is the second time I fix the same error in your code ( https://github.com/toggl/mobile/issues/656)

Phoebe/Data/Utils/TimeEntryGroup.cs
```diff
@@ -20,11 +22,29 @@ public class TimeEntryGroup : ITimeEntryModel
...
             Add (data);
         }
 
+        public TimeEntryGroup ()
+        {
+
+        }
+
+        public async Task BuildFromGuids (List<Guid> guids)
+        {
```
monday8am: @masterrr 
- Constructor overload is more elegant.

Phoebe/Data/Utils/TimeEntryGroup.cs
```diff
@@ -20,11 +22,29 @@ public class TimeEntryGroup : ITimeEntryModel
...
+            var first = new TimeEntryModel (guids.First ());
+            await first.LoadAsync ();
+            Add (first.Data);
+            foreach (var guid in guids.Skip (1)) {
+                var mdl = new TimeEntryModel (guid);
+                await mdl.LoadAsync ();
+                UpdateIfPossible (mdl.Data);
+            }
+        }
+
```
monday8am: - Avoid the creation of Models. If you create a new TimeEntryModel, you're filling a whole object with client, tasks, projects and you'll receive events about changes, etc. Is the second time I fix the same error in your code ( https://github.com/toggl/mobile/issues/656)

Phoebe/Data/Utils/TimeEntryGroup.cs
```diff
@@ -20,11 +22,29 @@ public class TimeEntryGroup : ITimeEntryModel
         private readonly List<TimeEntryData> dataObjects = new List<TimeEntryData> ();
         private TimeEntryModel model;
 
+
         public TimeEntryGroup (TimeEntryData data)
         {
             Add (data);
         }
 
+        public TimeEntryGroup ()
```
monday8am: The parameterless constructor is wrong, this object (TimeEntryGroup) make no sense if it is empty: if you request for any data you'll generate an exception.

Phoebe/Data/Extensions/ListExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+                    Guid.TryParse (guidEntity, out guidParsing);
+                    if (guidParsing != Guid.Empty) {
+                        ids.Add (guidParsing);
+                    }
+                }
+            }
+            return ids;
+        }
+    }
+}
```
monday8am: This class and the unique static method is associated all the time to the creation of a TimeEntryGroup. Pass a string[] to the overloaded constructor of TimeEntryGroup and don't use an extra class. Again, less code means less errors.

Phoebe/Data/Utils/TimeEntryFactory.cs
masterrr: I use the same code in more than one place, moved it out to the simple factory

monday8am: Yes, I understand and the general idea of the PR is ok :+1: 

But we need to go an step forward and try to move the code to Phoebe. I'll take care of the PR and try to show you some mainlines (later we can discuss it more deeply).

masterrr:
@monday8am thanks for comments, all fixed now :)

One more question: if we're not using the factory, how do I enforce DRY approach there?
https://github.com/toggl/mobile/pull/667/files#diff-cb6a2e616f96e832cd966f74587c4978R69
https://github.com/toggl/mobile/pull/667/files#diff-5e858785b885aac670344796607f9eedR38
https://github.com/toggl/mobile/pull/667/files#diff-5e858785b885aac670344796607f9eedR38

masterrr:
I made a static method in TimeEntryModel that allows us to build ITimeEntryModel instances

monday8am:
superseded by https://github.com/toggl/mobile/pull/686

Issue #639, opened 2015-05-02
Deleted item with Undo is added again if the app refresh items.


monday8am:
Add a Guid comprobation.

masterrr:
Fixed by https://github.com/toggl/mobile/pull/670

PR #670, opened 2015-05-13
Fix bulk entries deleting, time entry cell UI when switching activities

masterrr:
1. Fixes cell disappearing while holding swipe and e.g. pressing start button, then closing the new activity
2. Fixes removing 2+ entries in a row (the last one in the queue is not deleted and being nullified due to no guid check after the await). Fixing https://github.com/toggl/mobile/issues/639 as far as I get the issue & tested it. Both for single / grouped entries.

### DIFF
--- a/Joey/UI/Views/ListItemSwipeable.cs
+++ b/Joey/UI/Views/ListItemSwipeable.cs
@@ -89,6 +89,10 @@ namespace Toggl.Joey.UI.Views
 
         private void OnScrollAnimationEnd (object sender, EventArgs e)
         {
+            if (WindowVisibility == ViewStates.Gone) {
+                InitSwipeDeleteBg ();
+                return;
+            }
             if (SwipeAnimationEnd != null) {
                 SwipeAnimationEnd (this, EventArgs.Empty);
             }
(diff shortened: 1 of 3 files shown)


monday8am:
@masterrr I understand. The current solution (not your PR) waits for the delete operation of the last item added to add a new item.  I think the more simple solution could be just move the await operation to an external method. 

```
 public void ConfirmItemRemove ()
        {
            if (removedItem != null) {
                // trigger async delete 
                DeleteTimeEntry (removedItem);
                removedItem = null;
            }
        }

  private async void DeleteTimeEntry (TimeEntryData data)
        {
                var model = new TimeEntryModel (data);
                await model.DeleteAsync();
        }
```

masterrr:
@monday8am the current solution doesn't wait for delete to add a new item (if by the “item” you mean i-var named “removedItem”) and that's what causing the issue (no await or similar is used when calling ConfirmItemRemove())

Though I like your solution, just commited it (for the grouped also)

monday8am:
and it works? I can't test right now.

masterrr:
@monday8am works, tested grouped / single view

monday8am:
great.

Issue #672, opened 2015-05-16
Start toolbar button disappears after switching to the another activity via drawer, then going back


masterrr:

monday8am:
maybe this is the normal behaviour:
https://github.com/toggl/mobile/issues/611

@masterrr to be sure, behaivor should be the same as production version

masterrr:
@monday8am just checked out the app from the Google Play. The start button _appears_ if we are going back from the e.g. settings.

Though In our current android-redesign state the button doesn't come back (#611 is about switching off the button in other fragments, and yeah it's fixed), that's what this PR is about

masterrr:
![settings_button_demo](https://cloud.githubusercontent.com/assets/720966/7667538/874cad8c-fc14-11e4-946e-e11d2a7519f2.gif)

monday8am:
@masterrr :+1:

masterrr:
Closed by https://github.com/toggl/mobile/pull/677

PR #677, opened 2015-05-18
Always Show Timer Button In Toolbar If Main Timer Screen

masterrr:
The OpenPage() is called from the OnBackPressed(), though the method didn't care about timer showing if back to Main Timer screen
Fixing https://github.com/toggl/mobile/issues/672


taavirehemagi:
:+1:

PR #682, opened 2015-05-19
Splash screen implementation.

### LINKED ISSUE #640: Set a new Splash view
Splash view isn't showed.

taavirehemagi:
Closes #640


monday8am:
@taavirehemagi :+1: 
Just a question, do you know why we didn't have splash screen before? Probably there is a reason :)

taavirehemagi:
I'm not sure if there is a reason. This is the suggested way of doing it and I can see no reason why not to do it. Maybe it wasn't considered necessary before?

http://developer.xamarin.com/guides/android/user_interface/creating_a_splash_screen/

monday8am:
@taavirehemagi "splashdrawable" file is missing.

taavirehemagi:
rready to use

Issue #674, opened 2015-05-18
Shadows for ProjectList screen toolbar


monday8am:
- Add item shadow decorator to recyclerview.
- Add Toolbar shadow.

masterrr:
Closed by https://github.com/toggl/mobile/pull/685

PR #685, opened 2015-05-19
Project Screen: Toolbar, Last Project Cell Shadows

masterrr:
Closing https://github.com/toggl/mobile/issues/674
screenshot_2015-05-19-21-39-41


monday8am:
@masterrr is a bit tricky, later we should find a more elegant solution.

Issue #717, opened 2015-05-26
Project List Screen Loading Indicator (like in TE list)


masterrr:

masterrr:
Done by https://github.com/toggl/mobile/pull/721

PR #721, opened 2015-05-26
Recycled DataView Adapter: Display Spinner If Loading (affects Project Screen)

masterrr:
Closing https://github.com/toggl/mobile/issues/717
To eliminate “fragile base class” problem: only ProjectListAdapter, LogTimeEntriesAdapter and GroupedTimeEntriesAdapter inherit from the RecycledDataViewAdapter. These 3 are fully tested. 
screenshot_2015-05-26-23-54-57


monday8am:
explanation is more large than code fixed :P

Issue #719, opened 2015-05-26
Signup screen


pe0ny:
![toggl-android-signup-active-fields](https://cloud.githubusercontent.com/assets/5064445/7809135/d2d44740-03a0-11e5-9c74-11b70cdc1a61.png)

Active fields: 
* Active line - #f30c16 70%
* Active type - #f30c16 50%

Disclaimer: Roboto Regular (links Roboto Bold) 12sp #f5f5f5

Disabled button (#3b3b3b):
![toggl-android-signup-disabled-btn](https://cloud.githubusercontent.com/assets/5064445/7809174/18a4477a-03a1-11e5-9947-c025e9b41f62.png)

taavirehemagi:
Closed in #744

Issue #681, opened 2015-05-19
Sign up with google link appearance


taavirehemagi:

taavirehemagi:
Closed in #744

Issue #680, opened 2015-05-19
Make switch from login to signup form work with a link


taavirehemagi:

taavirehemagi:
Closed in #744

Issue #679, opened 2015-05-19
Login and signup controls design


taavirehemagi:
EditText and buttons look and behaviour

taavirehemagi:
Closed in #744

PR #723, opened 2015-05-27
New login screen design implementation

### LINKED ISSUE #678: Login and signup material design placement
toggl-android-signup

taavirehemagi:
Closes #678 #679 #680 #681 #719


monday8am:
@taavirehemagi  ..ehmm.. 
- the Google button is missing.
- margins are not correct (laterals should be 32dp @pe0ny)
- I'm not sure but the bg photo looks a bit deformed?

nexus 4 kitkat screentshot 3

taavirehemagi:
@monday8am  The google login only appears when there's accounts linked to google on the phone (just like the old version). But I agree, now looking at it, the legal text should be more at the bottom when it's missing. I'll check the bg deformation and the paddings, though.

taavirehemagi:
It appears that on android, backgrounds always stretch to match the container. For me, the screen was right enough size not to notice it before. Did a hack that's supposed to un-stretch the image, though. Can you validate, that it works for you, @monday8am? In addition, I fixed the margins as well.

monday8am:
in case of background, user a bitmap drawable as a background, then you can define image behaviour with the gravity attribute:
https://github.com/toggl/mobile/blob/android-redesign/Joey/Resources/drawable/BgArrowsPattern.xml
http://developer.android.com/reference/android/graphics/drawable/BitmapDrawable.html

monday8am:
wow, you did it :D , I'll check :+1:

Issue #701, opened 2015-05-20
Improve ShadowItemDecorator with a more elegant solution


monday8am:

masterrr:
@monday8am what do you think about the decorator? Should I try to improve the decorator, or take a different approach (like including shadows in a cell layout at the bottom)?

masterrr:
Closed by https://github.com/toggl/mobile/pull/726

Issue #699, opened 2015-05-20
Shadows on groups of items should be in both sides of material blocks.


monday8am:
Shadows should appears in both sides (top and bottom). Top shadow should be shorter.

masterrr:
Closed by https://github.com/toggl/mobile/pull/726

PR #726, opened 2015-05-27
Project List, Time Entry Log Screens: Shadows ❏ ❒

masterrr:
Closing https://github.com/toggl/mobile/issues/701 (gently applying shadows on the divider, so one type can handle bottom and top shadows at the same time)
Closing https://github.com/toggl/mobile/issues/699
screenshot_2015-05-27-21-30-51
screenshot_2015-05-27-21-31-28


monday8am:
:+1:

monday8am:
@masterrr two tips, 
- Currently the shadow decorator tries "to knows everybody" and decides if to put a shadow or not depending on the class and info passed (top or bottom). I think a more clean approach  is to create a custom attribute and decorate header classes and normal classes. Then you can check for the class attribute and decide.
http://edn.embarcadero.com/article/30250

- Is better to draw both shadows in the same decorator in terms of performance.

masterrr:
@monday8am done 🙌

PR #760, opened 2015-06-08
Disable ShowWhenLocked flag and DisableKeyguard permission.

### LINKED ISSUE #725: Find an alternative to ShowWhenLocked flag to make widget work.
On android 5.0 if the screen is locked, toggl app will stay on top of the lock screen. If to remove the flag, widget item clicks will not open app the way they are supposed to.

taavirehemagi:
App is brought to front when an entry is started and when the phone is unlocked the app comes to focus.
Closes #725


monday8am:
The flag has been removed because users are confused if the app is over the "screen guard". We are going to receive feedback with the flag removed and take a final decision.

monday8am:
The flag has been removed because users are confused if the app is over the "screen guard". We are going to receive feedback with the flag removed and take a final decision.

PR #762, opened 2015-06-08
Fix RecyclerView holder position error on continue

### LINKED ISSUE #745: Error on recyclerview
System.ArgumentOutOfRangeExceptionArgument is out of range.

```
at System.Linq.Enumerable.ElementAt<object> (System.Collections.Generic.IEnumerable`1<object>,int) <0x000f4>
at Toggl.Joey.UI.Adapters.RecycledDataViewAdapter`1<object>.GetEntry (int) <0x00077>
at Toggl.Joey.UI.Adapters.LogTimeEntriesAdapter.GetItemViewType (int) <0x00053>
at Toggl.Joey.UI.Adapters.LogTimeEntriesAdapter.BindHolder (Android.Support.V7.Widget.RecyclerView/ViewHolder,int) <0x00027>
at Toggl.Joey.UI.Adapters.LogTimeEntriesAdapter.OnContinueTimeEntry (Toggl.Phoebe.Data.DataObjects.TimeEntryData) <0x001a7>
at Toggl.Joey.UI.Adapters.LogTimeEntriesAdapter/TimeEntryListItemHolder.OnTouch (Android.Views.View,Android.Views.MotionEvent) <0x000bf>
at Android.Views.View/IOnTouchListenerInvoker.n_OnTouch_Landroid_view_View_Landroid_view_MotionEvent_ (intptr,intptr,intptr,intptr) <0x0007f>
at (wrapper dynamic-method) object.fb11cb30-ff59-4b34-a796-6f3c7d27456b (intptr,intptr,intptr,intptr) <0x0004b>
```

taavirehemagi:
holderList contains duplicate running item when changing from GroupedEntries mode with a running time entry.
Also, GetViewHolder function fills holderList up with cluttered data on continueing entries.
Closes #745

### DIFF
--- a/Joey/UI/Adapters/LogTimeEntriesAdapter.cs
+++ b/Joey/UI/Adapters/LogTimeEntriesAdapter.cs
@@ -101,10 +101,10 @@ namespace Toggl.Joey.UI.Adapters
 
             // Trick on view to show a better
             // visual reaction to press Play btn
-            for (int i = 0; i < holderList.Count; i++) {
-                var holder = holderList [i] as TimeEntryListItemHolder;
+            foreach (var item in holderList) {
+                var holder = item as TimeEntryListItemHolder;
                 if (holder != null) {
-                    if (holder.DataSource.State == TimeEntryState.Running) {
+                    if (holder.DataSource.State == TimeEntryState.Running && holder.AdapterPosition != -1) {
                         holder.DataSource.TimeEntryData.State = TimeEntryState.Finished;
                         BindHolder (holder, holder.AdapterPosition);
                     }


monday8am:
Totally agree. The bug is exactly that and is caused because the RecyclerView implementation 22.2.1.1 don't call this method:

```
        public override void OnViewDetachedFromWindow (Java.Lang.Object holder)
        {
            if (holder is TimeEntryListItemHolder) {
                var mHolder = (TimeEntryListItemHolder)holder;
                mHolder.DisposeDataSource ();
            } else if (holder is HeaderListItemHolder) {
                var mHolder = (HeaderListItemHolder)holder;
                mHolder.DisposeDataSource ();
            }
            base.OnViewDetachedFromWindow (holder);
        }
```

this method should clean the DataSource every time a holder is cached but is not called by the adapter.

monday8am:
There is an event in RecyclerView that could be called to dispose the DataSource for the cached holders:

```
            Owner.RecyclerEvent += (object sender, RecyclerView.RecyclerEventArgs e) => {
                // Seems to be the recycled holder
                e.Holder
            };
```

We should test to pass this holder to the previous posted method and see if the DataSource is disposed. In case of yes, we could use both solutions.

monday8am:
We'll merge this solution but we're waiting for the next update of RecyclerView to implement the correct solution (based on recyclerview methods)

PR #764, opened 2015-06-09
TE Activity: close if no data is restored from the intent

masterrr:
Fixing:
```
System.NullReferenceExceptionObject reference not set to an instance of an object
  at Toggl.Joey.UI.Fragments.EditTimeEntryFragment..ctor (Toggl.Phoebe.Data.DataObjects.TimeEntryData timeEntry) [0x00000] in <filename unknown>:0 
  at Toggl.Joey.UI.Activities.EditTimeEntryActivity+<OnCreateActivity>c__async0.MoveNext () [0x00000] in <filename unknown>:0 
```

Tested with "don't keep activities" and app foreground / background states

### DIFF
--- a/Joey/UI/Activities/EditTimeEntryActivity.cs
+++ b/Joey/UI/Activities/EditTimeEntryActivity.cs
@@ -31,6 +31,9 @@ namespace Toggl.Joey.UI.Activities
             TimeEntryData timeEntry = null;
             if (timeEntryList.Count > 0) {
                 timeEntry = timeEntryList[0];
+            } else {
+                OnBackPressed ();
+                return;
             }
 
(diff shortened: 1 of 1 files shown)


monday8am:
:+1: 
@masterrr you could simulate the error and then test the fix?
For this release is ok but we should discover the real cause of this bug.

masterrr:
@monday8am I simulated the count=0 list and got exactly this exception

masterrr:
@monday8am not sure about why no data is being fetched from the intent

monday8am:
@masterrr I think we can't close the issue if we don't know where the error comes.

monday8am:
@masterrr I'll open an issue to investigate the cause and close this one.

PR #769, opened 2015-06-14
Avoid wrong index on CanDissmis from swipe listener.

monday8am:
Fixed wrong on index on CanDismiss method that come from SwipeDismissTouchListener.


taavirehemagi:
:+1:

masterrr:
Good for me :)

PR #771, opened 2015-06-14
Review of group and continuous mode. Async operations and refactoring.

monday8am:
An extensive review of the update mechanism with the following improvements:
- Asynchronous methods to update entries.
- Revision of grouped methods add, delete and update. 
- Common holder for both collections.
- Common base class for both collection classes.

Simple class diagram.
chart2

Message processing flow.
chart1


Phoebe/Data/Utils/TimeEntryHolder.cs
```diff
@@ -0,0 +1,250 @@
...
+                projectData = await GetProjectDataAsync (timeEntryData.ProjectId.Value);
+                if (projectData.ClientId.HasValue) {
+                    clientData = await GetClientDataAsync (projectData.ClientId.Value);
+                }
+            }
+
+            if (timeEntryData.TaskId.HasValue) {
+                taskData = await GetTaskDataAsync (timeEntryData.TaskId.Value);
+            }
+            numberOfTags = await GetNumberOfTagsAsync (timeEntryData.Id);
```
masterrr: “The await operator is applied to a task in an asynchronous method to suspend the execution of the method until the awaited task completes”
Idea: rather than step-by-step await, decide what Get* methods are needed and then batch-call them so they run simultaneously -> faster 
@monday8am do you like the idea? I can profile if needed

monday8am:
batch-call  = put them in a list and use a Task.WaitForAll?

masterrr:
@monday8am yep, you tested it already? :)

monday8am:
no no, go for it if you have time.

masterrr:
@monday8am :+1:

PR #786, opened 2015-06-16
Project without tasks selection fixed.

monday8am:


monday8am:
@masterrr seems to be you were who defined this code. Please, check my code (maybe you can find a more elegant solution) and be careful with names to avoid confusions. (TaskHandler, ProjectHandler.. or things like that)

monday8am:
superseded by #797

Issue #772, opened 2015-06-15
Investigate bug related with PR #764


monday8am:
The bug is fixed but we should know why the data isn't correct.

masterrr:
I investigated this while fixing the bug https://github.com/toggl/mobile/pull/764
I couldn't find any meaningful reason of why it happens. 

EditTimeEntryFragment takes one param in the constructor: TimeEntryData entity. The Fragment is being instantiated from the Activity. The TimeEntryData should come to Activity from the intent, but rarely  it doesn't.

I can't reproduce intent extra = null both on devices & simulators so I force-simulated extra = null in code, gives me the exact same error. And I fixed it.

masterrr:
The last possible “reason” can be this, very similar issue http://stackoverflow.com/a/18154106/1279330 and we do it the wrong way, if projecting our solution

masterrr:
@monday8am I can try SO advice and we'll see if the issue persists

masterrr:
Goes pending via https://github.com/toggl/mobile/pull/793

monday8am:
fixed.

PR #793, opened 2015-06-19
Edit TE Activity: Bypass intent.Extras

masterrr:
Closing https://github.com/toggl/mobile/issues/772
http://stackoverflow.com/a/18154106/1279330

### DIFF
--- a/Joey/UI/Activities/EditTimeEntryActivity.cs
+++ b/Joey/UI/Activities/EditTimeEntryActivity.cs
@@ -49,14 +49,13 @@ namespace Toggl.Joey.UI.Activities
 
         public async static Task<IList<TimeEntryData>> GetIntentTimeEntryData (Android.Content.Intent intent)
         {
-            var extras = intent.Extras;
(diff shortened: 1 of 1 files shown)


monday8am:
the new code is better because there is a null check in the original call to getStringArrayListExtra
```
  public ArrayList<String> getStringArrayListExtra(String name) {
        return mExtras == null ? null : mExtras.getStringArrayList(name);
   }
```

but the issue is still open.  This is the reason of the trick's author: "i'm sure i've had the same issue and the answer i gave you did the trick and never heard of it again..." . Not enough for us.

masterrr:
@monday8am can I put “planned” tag on the issue and see if the bug persist after the update via Bugsnag?

monday8am:
@masterrr "planned" is another concept I think, more related with future improvements. Use "pending" tag ( I didn't created it, was you?) I have defined the same color as XI background :P

masterrr:
@monday8am “pending” is :+1: (yep, I created it :P)

Issue #785, opened 2015-06-16
SQLite bug


monday8am:
XI: 8a185675667a461552b697a8a372934a99a2f0bd
System.InvalidCastExceptionCannot cast from source type to destination type.

```
at (wrapper castclass) object:__castclass_with_cache (object,intptr,intptr)
  at SQLite.SQLiteCommand+<ExecuteDeferredQuery>c__Iterator0`1[Toggl.Phoebe.Data.DataObjects.UserData].MoveNext () [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[Toggl.Phoebe.Data.DataObjects.UserData].AddEnumerable (IEnumerable`1 enumerable) [0x00000] in <filename unknown>:0 
  at System.Collections.Generic.List`1[Toggl.Phoebe.Data.DataObjects.UserData]..ctor (IEnumerable`1 collection) [0x00000] in <filename unknown>:0 
  at System.Linq.Enumerable.ToList[UserData] (IEnumerable`1 source) [0x00000] in <filename unknown>:0 
```

monday8am:
Pending of check. #794

monday8am:
Closed in version 8.0.2

PR #794, opened 2015-06-19
SQLiteConnection: FullMutex (Serialized mode analog), put/delete locks

masterrr:
Issue: https://github.com/toggl/mobile/issues/785
I think the issue reason is multiple threads accessing SQLite store at once, fullMutex flag and base locks _should_ fix it. 
@monday8am @taavirehemagi need your opinion
http://forums.xamarin.com/discussion/549/sqlite-net-and-multiple-threads

### DIFF
--- a/Phoebe/Data/SqliteDataStore.cs
+++ b/Phoebe/Data/SqliteDataStore.cs
@@ -241,13 +241,15 @@ namespace Toggl.Phoebe.Data
             public Context (SqliteDataStore store, string dbPath)
             {
                 this.store = store;
-                conn = new SQLiteConnection (dbPath);
+                conn = new SQLiteConnection (dbPath, SQLiteOpenFlags.Create | SQLiteOpenFlags.ReadWrite | SQLiteOpenFlags.FullMutex);
             }
 
             public T Put<T> (T obj)
             where T : new()
             {
-                conn.InsertOrReplace (obj);
+                lock (conn) {
+                    conn.InsertOrReplace (obj);
+                }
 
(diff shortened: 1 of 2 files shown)


monday8am:
Maybe our code has this error @masterrr (currently we have an important memory leak related with Sqlite) but here I think the case is more simple. The first problem with this code is this method don't returns a Task:

``` 
        private async void ReloadUser ()
        {
            if (User == null) {
                return;
            }
            var store = ServiceContainer.Resolve<IDataStore> ();
            var rows = await store.Table<UserData> ().QueryAsync (r => r.Id == User.Id);
            User = rows.FirstOrDefault ();
        }
``` 

If an async method don't return a task then you can't catch an exception inside the try-catch that appears in the constructor (https://msdn.microsoft.com/en-us/magazine/jj991977.aspx). And the problem probably is more simple, maybe there is no users in the database and rows = null and FirstOrDefault () generate an exception. Possible solution could be:

- Return a Task from ReloadUser.
- Create an Init () method that is called from constructor ( you can't decorate a constructor with async) and move methods from constructor to it.
- Await ReloadUser inside the try-catch.

The bug will be still there but it will be caught correctly and the rest of the app will work as expected. (I hope)

masterrr:
@monday8am please, check the AuthManager, if I got your idea right :)

monday8am:
nice! :+1: 
but the previous commit is still there. Edit the history (rebase -i and remove it, then force a push)

masterrr:
@monday8am done

PR #796, opened 2015-06-19
Remove RecentViewController

masterrr:


monday8am:
:+1: for me.

In some moment we'll try to get back the recent projects functionality (some people like it), the key is how :)

PR #797, opened 2015-06-22
Project List: Tasks Reveal Icon, Separate Handlers; Remove Obsolete Project Dialog

masterrr:
— Ability to choose project w/o tasks if project has some 
— New tasks reveal icon, instead of “tasks” label; remove vert. separator
— Remove ChooseTimeEntryProjectDialogFragment.cs, ProjectsAdapter.cs
nexus 5 lollipop screentshot 2


Joey/Resources/layout/ProjectListProjectItem.axml
```diff
@@ -16,7 +16,7 @@
         android:layout_height="wrap_content"
         android:layout_weight="1"
         android:layout_marginLeft="29dp"
-        android:layout_marginRight="32dp"
+        android:layout_marginRight="15dp"
```
monday8am: the values for margins are 16dp or 32dp in all screens. In this case should be 16dp (ask @pe0ny anyway)

monday8am:
@masterrr As a general rule, every button that you define should have a normal - pressed state (in case of this Task button, an arrow was defined). The ideal behaviour is like Clock app in Android 5, but maybe the animation is too much for this moment. You should add:

- ripple / background change when the user press the icon.
- selected state: darker color for the icon or arrow. ( I prefer solution 1, considering we have to implement a transition between normal icon and pressed icon, is more easy to change the color)

masterrr:
> selected state: darker color for the icon or arrow. ( I prefer solution 1, considering we have to implement a transition between normal icon and pressed icon, is more easy to change the color)
I'll vote for solution 1 too (darker color). As @pe0ny is not available, I'll generate darker resources myself. 

“ripple / background change when the user press the icon.” issue is still open, can't get it to work :mag:

masterrr:
@monday8am darker color is done

monday8am:
nice! first point is done:

- ripple / background change when the user press the icon. (OK)
- selected state: darker color for the icon or arrow. ( I prefer solution 1, considering we have to implement a transition between normal icon and pressed icon, is more easy to change the color)

second point? the button is dark until it is pressed again?

masterrr:
@monday8am tasks button is dark when pressed, and stays dark until it's un-pressed (collapsing tasks list)

monday8am:
great! I'll check then in a while.

Issue #776, opened 2015-06-16
iOS Push Notification Client Implementation


masterrr:

masterrr:
Closed by https://github.com/toggl/mobile/issues/801

Issue #775, opened 2015-06-16
iOS Push Notification Backend Scheme Check


masterrr:

masterrr:
Closed by https://github.com/toggl/mobile/issues/801

PR #801, opened 2015-06-29
iOS Push Notifications

masterrr:
Closing https://github.com/toggl/mobile/issues/776 https://github.com/toggl/mobile/issues/775

### DIFF
--- a/Ross/AppDelegate.cs
+++ b/Ross/AppDelegate.cs
@@ -84,7 +84,7 @@ namespace Toggl.Ross
 
         public override void DidReceiveRemoteNotification (UIApplication application, NSDictionary userInfo, System.Action<UIBackgroundFetchResult> completionHandler)
         {
-            Task.Run (() => APNSManager.DidReceiveRemoteNotification (application, userInfo, completionHandler));
+            Task.Run (() => APNSManager.DidReceiveRemoteNotificationAsync (application, userInfo, completionHandler));
         }
 
 
--- a/Ross/AppDelegate.cs
+++ b/Ross/AppDelegate.cs
@@ -14,9 +14,8 @@ using Toggl.Ross.Net;
 using Toggl.Ross.ViewControllers;
 using Toggl.Ross.Views;
 using UIKit;
-using XPlatUtils;
 using Xamarin;
-using System.Threading.Tasks;
+using XPlatUtils;
 
 namespace Toggl.Ross
 {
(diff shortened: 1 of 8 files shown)


monday8am:
@masterrr please, refactor the method name 

https://github.com/toggl/mobile/blob/master/Phoebe/Net/PushRestClient.cs#L60
https://github.com/toggl/mobile/blob/master/Phoebe/Net/PushRestClient.cs#L77

with the Async termination.

Ross/AppDelegate.cs
```diff
@@ -26,6 +26,7 @@ public partial class AppDelegate : UIApplicationDelegate, IPlatformInfo
         private TogglWindow window;
         private int systemVersion;
         private const int minVersionWidget = 7;
+        private bool componentsInitialized = false;
```
monday8am: we're not using this.

Ross/AppDelegate.cs
```diff
@@ -53,10 +54,27 @@ public override bool FinishedLaunching (UIApplication application, NSDictionary
             ServiceContainer.Resolve<IBugsnagClient> ();
             ServiceContainer.Resolve<BugsnagUserManager> ();
             ServiceContainer.Resolve<ITracker> ();
+            ServiceContainer.Resolve<APNSManager> ();
```
monday8am: if you are resolving this object there, save the reference and you don't have to resolve it in the event handlers.. (RegisteredForRemoteNotifications, FailedToRegisterForRemoteNotifications, etc)

Ross/Net/APNSManager.cs
```diff
@@ -0,0 +1,170 @@
...
+{
+    public class APNSManager
+    {
+        private static readonly string Tag = "APNSManager";
+
+        private const string PushDeviceTokenKey = "PushDeviceToken";
+        private readonly object subscriptionAuthChanged;
+
+        private string authToken;
+        private string SavedDeviceToken { get { return NSUserDefaults.StandardUserDefaults.StringForKey (PushDeviceTokenKey); } }
```
monday8am: use the usual property format.

monday8am:
to new little tasks:
- Check if is needed to do RegisterDevice and UnregisterDevice "awaitables" . Maybe you need to exec one and wait to exec the other one. In case of not, avoid to call them consecutively with the same token.
- Makes DidReceiveRemoteNotification "awaitable" (return a Task) and use try-catch like in Android version. Then await for it in the event handler (DidReceiveRemoteNotification).

the rest for me is ok :+1:

masterrr:
@monday8am check it out now :+1:

Issue #835, opened 2015-07-20
iOS app 7.022. rejected.


monday8am:
July 18, 2015 at 4:15 AM

From Apple
11.13 - Apps that link to external mechanisms for purchases or subscriptions to be used in the App, such as a "buy" button that goes to a web site to purchase a digital book, will be rejected
11.13 Details

We noticed that your app provides access to external mechanisms for purchases or subscriptions to be used in the app, which does not comply with the App Store Review Guidelines. 

Please remove the following:

- In the Application Description: www.toggl.com links to Safari
- In the Application Description: "* Pro account feature. You can upgrade from the web app." 
-In the binary, Account Creation: Terms and Conditions links to Safari 

We’ve attached screenshot(s) for your reference.

Next Steps

Please remove the access to these mechanisms - including fully qualified links to your site that could indirectly provide access to these mechanisms, such as links to web pages for support, FAQ, product or program details, etc.

Image 1:
![temp tmlegphs](https://cloud.githubusercontent.com/assets/1842773/8775979/c398ed58-2eed-11e5-92be-4d9a88d8cb99.png)

Image 2: 
![temp glzelkng](https://cloud.githubusercontent.com/assets/1842773/8775981/c6f86d7a-2eed-11e5-8c2c-dacb36045fbb.png)

monday8am:
@masterrr besides the new link, we need to open the page in an internal WebView

monday8am:
and... this is the issue from web app team to strip out of links the page
https://github.com/toggl/website/issues/281

masterrr:
Closed by https://github.com/toggl/mobile/pull/839

PR #839, opened 2015-07-21
Use an Internal WebView for TOS Links

masterrr:
download 2


monday8am:
Fixes #835 
@masterrr don't forget to reference the original issue.

masterrr:
Done!
download 4

Issue #864, opened 2015-08-10
Project selection screen overhaul


taavirehemagi:
![screen shot 2015-08-11 at 00 18 59](https://cloud.githubusercontent.com/assets/2154171/9184559/a19a1d80-3fbe-11e5-9920-b3de6a5b4bf1.png)
Layout:
- RecyclerView will hold one workspace worth of projects at a time.
- TabLayout will be used for switching workspaces (has to slide away on scroll)
- Floating Action Button will be used for adding projects.

Add Project dialog will get a "Add client" field.

taavirehemagi:
![select_project](https://cloud.githubusercontent.com/assets/2154171/9199740/ceb8846c-404c-11e5-9a7a-205d4e37fcaf.gif)

PR #896, opened 2015-08-26
Project selection view

taavirehemagi:
Closes https://github.com/toggl/mobile/issues/864


Joey/UI/Activities/ProjectListActivity.cs
```diff
@@ -1,36 +1,175 @@
...
+        private class ProjectFragmentAdapter : FragmentPagerAdapter
+        {
+            private List<Fragment> fragments = new List<Fragment> ();
+            private List<String> fragmentTitles = new List<String> ();
+            private List<WorkspaceData> workspaces;
+            private int firstTabPos = 0;
+
+            public ProjectFragmentAdapter (FragmentManager fm, IList<TimeEntryData> timeEntryList) : base (fm)
+            {
+                workspaces = WorkspaceProjectsView.GetWorkspaces().Result;
```
monday8am: This code "blocks" an async call. The order of how to create views and load data should be changed. The starting point is OnCreateActivity, then we should:

- Init native components.
- Parse data passed to activities.
- Load data from db using correct async calls.
- Feed native components with the data loaded.

Phoebe/Data/Views/WorkspaceProjectsView.cs
```diff
@@ -521,19 +512,45 @@ public bool IsLoading
             }
         }
 
+        public static async Task<List<WorkspaceData>> GetWorkspaces()
```
monday8am: GetWorkspaces should be GetWorkspacesAsync. 

This method is not used inside the class. Together with SortWorkspaces, the only purpose together with SortWorkspaces is to returns sorted workspaces to an external object. And it doesn't use any internal resource. 

We should move it to another place.

monday8am:
Considering what we talked before, there is a change of approach. My suggestion is:
- Keep only one ProjectListViewModel like before.
- WorkspaceProjectsView is inside the view model and should have a workspace modifier. If you pass a workspace id the content should be change.
- The workspace change should be triggered by **TabLayout.OnTabSelectedListener**

taavirehemagi:
For now, my solution would be do keep the approach and pager the same, just disable the swiping. In my opinion, this approach keeps the animation of views sliding and the models are more easy and smaller. You feel strongly about having one ProjectListModel? Workspace-based is good for now in my opinion, since we're not sure how it will look/be used in iOS. 

Also, can you check the state now? I think I fixed what you told me.

monday8am:
@taavirehemagi 
> You feel strongly about having one ProjectListViewModel
We must have a ProjectListViewModel because is the only way it makes sense. In the ViewModel we should have represented everything that appears in the view. 

- If you keep the viewPages, then you have to keep different lists. This is more complex than one list.
- And you show to the users that the lists can be moved (right - left). But they can't moved them. Is confused. Is the problem of use half functionality in a component.

taavirehemagi:
aalright :+1:

taavirehemagi:
I'm running out of clever things to do here... :) Want to check? @monday8am

Issue #874, opened 2015-08-13
Bug in iOS related with the loading of Tags


monday8am:
Under certains conditions, the load of tags seems to be that breaks the Edit Entry screen.
(862198)

- Open app.
- Select one time entry or start one.
- Loader appears, no tags are loaded.
- App is frozen for 20 and then crash.

masterrr:
More info: load appears, frozen for ~5–10 secs, then tags are loaded and displayed

masterrr:
The issue: trie add operation is the bottleneck, blocking main thread https://github.com/toggl/mobile/blob/059a8638f8379924237f6a2f01bf429032973146/Phoebe/Data/Views/SuggestionEntriesView.cs#L84

masterrr:
The reason is TRIE can't handle long strings. Even `entry.Description.Substring(0, 20)` solves this (20 symbols length is really enough for the autocompletion). 

You can notice performance going down after the description string hits 200+ length treshold. Users love long descriptions :yum: 

Some TRIE projects advice to remove long strings (1000+ length) to increase the performance
http://www.cs.berkeley.edu/~jnwang/projects/triejoin/

Investigating further...

masterrr:
Don't close, let's watch how the fix behaves in production?

monday8am:
@masterrr with the user I passed you, the crash occurred when the user displayed the EditTimeEntryViewController. I was thinking that TRIES only is activated when the user starts to type.

masterrr:
SuggestionViewController is only actived when user starts to type, but... trie generation happens behind the screen when Edit VC is revealed

monday8am:
we should move trie generation to a background thread. **Not now** but we should move all the functionality to the view model and encapsulate it inside an async Task.

monday8am:
Closed.

PR #899, opened 2015-08-26
Autocompletion: limit TRIE string node max length

masterrr:
Closing https://github.com/toggl/mobile/issues/874 (see bug explanation there)


monday8am:
@masterrr did you tested that the bug appears/ disappear changing the char number?

masterrr:
@monday8am yes, if you change it to ~50–100 you can see some main thread blocked for a moment, set it to 1000+ and it's blocked "forever"  (ofc if you have descriptions _that_ long)

Issue #909, opened 2015-08-31
Rebase dmitry-group-ios-option1 w/ master


masterrr:

masterrr:
Closed by https://github.com/toggl/mobile/tree/dmitry-te-viewmodel-m

PR #913, opened 2015-09-01
ViewModels Integration Across View Controllers

masterrr:
Issue https://github.com/toggl/mobile/issues/909


monday8am:
out of date.

PR #962, opened 2015-10-15
Updated multiple libs including Android Support, Google Services and XI.

monday8am:
Updated many Nuget packages.


taavirehemagi:
Looks good to me!

PR #979, opened 2015-10-23
Libs related with mvvm.

monday8am:


monday8am:
@taavirehemagi only libs (as you requested)

PR #986, opened 2015-10-28
LogTimeEntryListViewModel PR

monday8am:


Joey/UI/Components/TimerComponent.cs
```diff
@@ -120,6 +138,38 @@ private bool SyncModel ()
...
+        public EventHandler FABStateChange;
+
+        public FABButtonState EntryState
+        {
+            get {
+                return entryState;
+            }
+        }
+
+        private void SendState ()
```
taavirehemagi: We don't need this function anymore, since the FAB in Log entries list updates itself. Also, we don't need the ActiveEntryChanged event.

monday8am: :+1:

monday8am:
closed. will be sent again against master.

Issue #987, opened 2015-10-29
Projects in most used view don't have Tasks icon.


taavirehemagi:

taavirehemagi:
Closed by #1001

PR #1001, opened 2015-11-03
Taavi searchfixes

### LINKED ISSUE #988: Search in projects should also go through clients and tasks

taavirehemagi:
Closes #988 #987

### DIFF
--- a/Joey/UI/Fragments/ProjectListFragment.cs
+++ b/Joey/UI/Fragments/ProjectListFragment.cs
@@ -255,7 +255,7 @@ namespace Toggl.Joey.UI.Fragments
 
         private void SearchList()
         {
-            if (string.IsNullOrEmpty (filter)) {
+            if (filter == null) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


Joey/UI/Fragments/ProjectListFragment.cs
```diff
@@ -255,7 +255,7 @@ public bool OnQueryTextChange (string newText)
 
         private void SearchList()
         {
-            if (string.IsNullOrEmpty (filter)) {
+            if (filter == null) {
```
monday8am: why you prefer this way?

taavirehemagi: if the search string goes from "a" -> "", then it is supposed to display the no-filter state of the list but since string.IsNullOrEmpty("") == true, then it gets cut off and you can't get back to the original state of the list when you delete the last letter.

monday8am: :+1:

monday8am:
if it works, :+1: ready to merge.
Not now but in some moment I'll put my hands over this class: *WorkspaceProjectsView.cs* :)

taavirehemagi:
It works, shall I merge myself or will you?

PR #1003, opened 2015-11-04
Update to multiples ViewModels.

monday8am:
The follow VM were updated:
- NewProjectViewModel
- ClientListViewModel
- CreateClientViewModel

Now the structure is more adaptive to change. The dialogs only return a ClientData object and there are not "Models" objects except inside the ViewModels. 
- CreateClientDialog: Create a client and returns a ClientData (doesn't modify a Project)
- ClientListDialog: Only returns a ClientData.
- All operations are made in NewProjectViewModel ( or the corresponding fragment).

To pass data between dialogs, a listener-interface pattern is used, more like Android style.


taavirehemagi:
Looks decent to me! Good progress.

PR #1011, opened 2015-11-16
OBM first test implementation

taavirehemagi:
- Only send "click" action, when the Start button was pressed with empty recyclerView (this includes users WITH experiment view and users with NO empty state)
- Only show empty state for users that have ExperimentNumber == 75 and are in the included group.


Phoebe/ExperimentNumbers.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Toggl.Phoebe
+{
+    public static class ExperimentNumbers
+    {
+        public const int HomeEmptyState = 75;
+    }
+}
+
```
monday8am: We could put this info inside ExperimentAction as static constants..
`var action = ExperimentAction.HomeEmptyState; ` 
or as an public enum inside this file. 
`var action = ExperimentAction.ExperimenIds.HomeEmptyState` .. the key is that they are always together.

Other detail, the name is ExperimentNumbers and the target property is ExperimentId (numbers or Ids?).

Phoebe/Phoebe.Android.csproj
```diff
@@ -77,6 +77,8 @@
     <Compile Include="Threading\**\*.cs" />
     <Compile Include="Analytics\**\*.cs" />
     <Compile Include="Logging\**\*.cs" />
+    <Compile Include="Data\Json\ActionJson.cs" />
+    <Compile Include="Net\ExperimentAction.cs" />
   </ItemGroup>
   <ItemGroup>
```
monday8am: we don't need this change in *Phoebe/Phoebe.Android.csproj*. It happens to me all the time :)

Joey/UI/Fragments/LogTimeEntriesListFragment.cs
```diff
@@ -83,7 +85,21 @@ public override View OnCreateView (LayoutInflater inflater, ViewGroup container,
...
+            await ViewModel.StartStopTimeEntry ();
+
+            if (UserInExperimentGroups && ViewModel.HasMore) {
+                var experimentAction = new ExperimentAction () {
+                    ExperimentId = ExperimentNumbers.HomeEmptyState,
+                    ActionKey = "startButton",
+                    ActionValue = "click"
+                };
+                await experimentAction.Send ();
+            }
```
monday8am: we need to move this logic to Phoebe. The condition related with the View in this case is *ViewModel.HasMore*, the rest of logic belongs to Phoebe. The line of code should be a call like:
`ExperimentManager.Send (ExperimentManager.HomeEmptyState, "startButton", "click");`

we don't need to await it because it could happen in the background. *ExperimentManager* is an example but the final idea is to be similar to Google Analytics or to Xamarin.Insights. Something like a service that you call and pass three parameters and forget.

taavirehemagi: :+1:  Will do

taavirehemagi:
Is it better now?

monday8am:
yeeahh!

Phoebe/Net/TogglRestClient.cs
```diff
@@ -810,6 +813,19 @@ private IEnumerable<TimeEntryJson> GetChangesTimeEntryObjects (JToken json, User
...
+        {
+            var url = new Uri (v9Url, "obm/actions");
+            var json = JsonConvert.SerializeObject (jsonObject);
+
+            var httpReq = SetupRequest (new HttpRequestMessage () {
+                Method = HttpMethod.Post,
+                RequestUri = url,
+                Content = new StringContent (json, Encoding.UTF8, "application/json")
+            });
+            var response = await SendAsync (httpReq).ConfigureAwait (true);
```
monday8am: .ConfigureAwait (false);

Phoebe/OBMExperimentManager.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public static async void Send (int experimentNumber, string actionKey, string actionValue)
+        {
+            if (InExperimentGroups (experimentNumber)) {
+                var experimentAction = new ExperimentAction () {
+                    ExperimentId = experimentNumber,
+                    ActionKey = actionKey,
+                    ActionValue = actionValue
+                };
+                await experimentAction.Send ();
```
monday8am: `Task.New (async () => {
await experimentAction.Send ();
}`

taavirehemagi:
Fixed!

monday8am:
nice!

PR #1021, opened 2015-11-18
Fix build scripts

alfonsogarciacaro:
Update FAKE to 4.4.2
Fix minor issues in F# build scripts


monday8am:
let see if it works ok.

PR #1027, opened 2015-11-23
Swipe refresh for home screen

### LINKED ISSUE #1026: Replace drawer sync button with swipe refresh for log entries screen

taavirehemagi:
To refresh, just scroll the recyclerview up. Status of last sync is still showed in the old place @ drawer.

Closes #1026

### DIFF
--- a/Joey/UI/Fragments/LogTimeEntriesListFragment.cs
+++ b/Joey/UI/Fragments/LogTimeEntriesListFragment.cs
@@ -4,6 +4,7 @@ using Android.Content;
 using Android.OS;
 using Android.Support.Design.Widget;
 using Android.Support.V4.App;
+using Android.Support.V4.Widget;
 using Android.Support.V7.Widget;
 using Android.Support.V7.Widget.Helper;
 using Android.Views;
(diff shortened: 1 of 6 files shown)


monday8am:
Ready to merge for me. I added a notification to show only errors. If you both like it, we can kept it. 
To test, go to airplane mode and do a pull&request (google apps do the same)

Joey/UI/Fragments/LogTimeEntriesListFragment.cs
```diff
@@ -202,6 +218,40 @@ public bool CanClick (RecyclerView view, int position)
 
         #endregion
 
+        public void OnRefresh ()
+        {
+            // Sync manager doesn't launch a sync op if a Time Entry
+            // is running
+            if (ViewModel.IsTimeEntryRunning) {
+                swipeLayout.Refreshing = false;
+                return;
```
alfonsogarciacaro: I'm testing the code in my phone but whenever I try to sync with an entry running, the code gets into this block and returns from the method. Thus, `syncManager.Run()` below is not called and there's no actual syncing.

monday8am: my mistake, I confused the Running in `SyncManager` with the Running in `ActiveTimeEntryManager`. I'll remove this lines. :+1:

PR #1028, opened 2015-11-23
Mvvm and state refactor of Android views.

monday8am:
Big PR although with not a lot of content and 700 lines less. Two main advances:

- Mvvm: Now every View - ViewModel has the same structure and uses the Fody and Mvvm Light libraries to implement the INotifyProperty interface and to bind changes to the view. This structure will be described in Wiki.

- State: Some months ago, the structure of the app was the follow: a `TimeEntryModel` was modified in every screen: if you had to modify a time entry tags, you had to pass the `TimeEntryData` *Id*, instantiate a `TimeEntryModel`, load tags, and modify the `TimeEntryModel` and saved it. Besides that, this code was mixed with View code. The first step was to move this code to ViewModels, and the second step was to isolate and reduce functionalities to the views.
Now, a tag list view is only that, a tag list and this view doesn't modify a `TimeEntryData` or `TimeEntryModel`, it only returns a selected list of tags. The same for projects, a ProjectList view doesn't modify a TimeEntryData, only shows a list of projects returns one if it is selected. And `TimeEntryData` objs are only modified in a ViewModel called `EditTimeEntryViewModel`. 
This change reduce the amount of code and bugs. More details will be described in the repo Wiki.

- Manual form is introduced. The manual form is exactly the same view used to edit existing time entries, but with a *new* and *stopped* time entry.


Joey/Joey.csproj
```diff
@@ -721,6 +716,16 @@
     <AndroidResource Include="Resources\drawable-hdpi\IcManualMode.png" />
     <AndroidResource Include="Resources\drawable-xhdpi\IcManualMode.png" />
     <AndroidResource Include="Resources\drawable-xxhdpi\IcManualMode.png" />
+    <AndroidResource Include="Resources\menu\SaveItemMenu.axml" />
+    <AndroidResource Include="Resources\drawable-mdpi\ic_done_black_24dp.png" />
+    <AndroidResource Include="Resources\drawable-hdpi\ic_done_black_24dp.png" />
+    <AndroidResource Include="Resources\drawable-xhdpi\ic_done_black_24dp.png" />
+    <AndroidResource Include="Resources\drawable-xxhdpi\ic_done_black_24dp.png" />
+    <AndroidResource Include="Resources\menu\NewItemMenu.axml" />
+    <AndroidResource Include="Resources\drawable-hdpi\ic_add_black_24dp.png" />
```
taavirehemagi: We already have that resource in our project. The files are IcManualMode.png for + icon and IcCheck.png for the check icon. Also, the icon names should follow the naming convention we have been using so far.

taavirehemagi:
When adding a tag, the dialog closes and the newly created tag does not show up in tag input field. 
1) the tag should be  selected in list after the creation 
2) the tag list should not close, since the user may want to select some more tags or create another one.

Joey/Resources/values/Strings.xml
```diff
@@ -124,6 +124,7 @@
     <string name="ChooseTimeEntryTagsDialogTitle">Select tags</string>
     <string name="ChooseTimeEntryTagsDialogCancel">Cancel</string>
     <string name="ChooseTimeEntryTagsDialogOk">Ok</string>
+    <string name="ChooseTimeEntryTagsDialogCreate">Create New Tag!</string>
```
taavirehemagi: Since it's already in a tag dialog, the button should say "create new ".

monday8am: :+1:  .. in case of position, Neutral button at center is the position used for this cases. Sincerely I forgot to change it.

Joey/UI/Components/TimerComponent.cs
```diff
@@ -26,301 +12,77 @@ namespace Toggl.Joey.UI.Components
...
+            isRunningBinding = this.SetBinding (() => ViewModel.IsTimeEntryRunning, () => IsRunning);
+            durationBinding = this.SetBinding (() => ViewModel.Duration, () => DurationTextView.Text);
+            descBinding = this.SetBinding (() => ViewModel.Description, () => DescriptionTextView.Text)
+                          .ConvertSourceToTarget (desc => desc != string.Empty ? desc : activity.ApplicationContext.Resources.GetText (Resource.String.TimerComponentNoDescription));
+            projectBinding = this.SetBinding (() => ViewModel.ProjectName, () => ProjectTextView.Text)
+                             .ConvertSourceToTarget (proj => proj != string.Empty ? proj : activity.ApplicationContext.Resources.GetText (Resource.String.TimerComponentNoProject));
         }
 
-        public bool Hide
+        public void DetachBindind ()
```
taavirehemagi: typo here, DetachBinding I think? :)

Joey/Joey.csproj
monday8am: the name convention is the followed at: https://www.google.com/design/icons/
The icons are downloaded from this place and I'm not sure of our current icon sizes.
IcAdd has a size of 30x30 and ic_add_black_24dp has a size of 36x36. Every icon right now has 36x36 for hdpi.

taavirehemagi: Yeah, but all the other icons in our app are named in a different style. I think inside the project,  naming should be consistent. And also, if you replaced the current ones with the new ones, then should at least delete the current (or old) icons.

Joey/UI/Components/TimerComponent.cs
monday8am: 1) Bug maybe, I'll check it
2) I could say .. users maybe want to create a Tag but not to select it, just create. I just followed the current behaviour: tag list closes after you create one. If you want to change this behaviour, we could do it but with more opinions ;)

taavirehemagi: Right now, it actually selects it, but it doesn't show up on the first close. 
To test:
- create a tag
- the dialog closes
- open the dialog again, you see the tag is selected now
- close it again
- appears in input

atleast that's how it is for me

taavirehemagi:
The check icon I think is too big and entirely black compared to the arrow icon. Should be smaller and gray. Maybe ask heidi to help?
screen shot 2015-11-25 at 12 37 38

Joey/UI/Fragments/EditTimeEntryFragment.cs
```diff
@@ -52,120 +79,186 @@ public static EditTimeEntryFragment NewInstance (string timeEntryId)
...
+                               .SetName (Resource.String.BaseEditTimeEntryFragmentDescription);
+
+            ProjectField = view.FindViewById<TogglField> (Resource.Id.Project)
+                           .SetName (Resource.String.BaseEditTimeEntryFragmentProject)
+                           .SimulateButton();
+
+            TagsField = view.FindViewById<TogglTagsField> (Resource.Id.TagsBit);
+            BillableCheckBox = view.FindViewById<CheckBox> (Resource.Id.BillableCheckBox).SetFont (Font.RobotoLight);
+
+            DurationTextView.Click += (sender, e) =>
```
taavirehemagi: I'm not sure that anynomous functions should be used here, since you can't unsubscribe them?

monday8am: about readability, I don't have a defined opinion. I don't like too much a lot of methods with 3 lines, but is a question of taste. Make a PR fixing it and I'll add if you want :)

Joey/UI/Components/TimerComponent.cs
```diff
@@ -26,301 +12,77 @@ namespace Toggl.Joey.UI.Components
...
-                OpenTimeEntryEdit (ActiveTimeEntry);
-            }
-        }
+            ViewModel = viewModel;
 
-        private void CreateTimeEntryManually (object sender, EventArgs e)
-        {
-            OpenTimeEntryEdit (new TimeEntryModel());
+            // TODO: investigate why WhenSourceChanges doesn't work. :(
+            isRunningBinding = this.SetBinding (() => ViewModel.IsTimeEntryRunning, () => IsRunning);
```
alfonsogarciacaro: I suspect MvvmLight expects `this` to be the source which is not the case for `isRunningBinding`

monday8am: I still don't know.. here it works:
https://github.com/toggl/mobile/blob/f626b07f2e80bc6e7f0de305191a6ac78eb18cc3/Joey/UI/Fragments/LogTimeEntriesListFragment.cs#L78

alfonsogarciacaro: Interesting... there you don't define a target for the binding. May this be related to the issue?

Issue #1057, opened 2015-12-02
Confusing buttons in Client Select Dialog


alfonsogarciacaro:
In version 8.0.8 there are now two buttons in the Client Select Dialog reading "New client" and "Create". This is confusing for the users: "Create" should actually read "Ok"

<img width="410" alt="screen shot 2015-12-02 at 08 56 19" src="https://cloud.githubusercontent.com/assets/8275461/11527984/59d7f402-98e4-11e5-9560-7f4a6ed62c98.png">

monday8am:
for me, we need a couple of things here:
- Manage the case that there are not clients (show a label saying "No clients")
- Replace the "Create" by "Cancel" .. but in this case, I think maybe "Cancel" is not needed. 
@alfonsogarciacaro  @taavirehemagi

monday8am:
closed by: https://github.com/toggl/mobile/pull/1061

PR #1061, opened 2015-12-03
Fix issue #1057: Confusing buttons in Client Select Dialog

alfonsogarciacaro:
Just sending the PR so it can be reviewed, code can still change.
This PR is doing two things:
* Eliminate button "Create" in Select Client dialog. The function of this button was actually to dismiss the dialog, but this can be done by touching outside the dialog so it's not actually needed ("New client" button remains).
* When selecting a client in the Edit Project screen, if there're no clients, the New Client dialog will be open directly instead of the Select Client dialog.


taavirehemagi:
I think there still should be a Dismiss button as well, regardless of the fact that the fragment can be dismissed otherwise. So, two buttons, new client, and dismiss would be my way to go. What do you guys think?

monday8am:
@taavirehemagi we'll go for the simplest way (one button). If you want to add an experiment for one-two buttons, is a good idea.

PR #1085, opened 2015-12-08
Client screens linked to ViewModels

monday8am:
Linked to ViewModels the screens `NewClientViewController` and `ClientSelectionViewController`


Phoebe/Data/ViewModels/CreateClientViewModel.cs
```diff
@@ -20,11 +20,19 @@ public CreateClientViewModel (Guid workspaceId)
...
+        }
+
         public async Task Init ()
         {
             IsLoading = true;
 
-            workspaceModel = new WorkspaceModel (workspaceId);
+            if (workspaceModel == null) {
+                workspaceModel = new WorkspaceModel (workspaceId);
+            }
```
alfonsogarciacaro: If calling `Init` is always mandatory it is less prone to failure if the constructor is private and the `async` method is marked as static and returns the instance. This probably will also remove the need to have `IsLoading` as nobody will have a reference of the instance until it's already loaded:
```
CreateClientViewModel ()
{
    ServiceContainer.Resolve<ITracker> ().CurrentScreen = "New Client Screen";
}

public static async Task<CreateClientViewModel> Init (WorkspaceModel workspaceModel = null)
{
    var vm = CreateClientViewModel();
    vm.workspaceModel = workspaceModel ?? new WorkspaceModel (workspaceId);
}
```

monday8am: yeapp, we can apply this pattern.. I didn't do it because I'm lazy :)
http://blog.stephencleary.com/2013/01/async-oop-2-constructors.html

I'll fix it.

monday8am: Issue opened to change all viewModels at time:
https://github.com/toggl/mobile/issues/1086

Issue #1095, opened 2015-12-14
Setting related with "Ask for project" is not working.


monday8am:
Show project screen after start new entry.

taavirehemagi:
Closed in #1096

Issue #1088, opened 2015-12-09
Grouped mode should be disabled for new users by default


kyllikekimmel:
Right now it looks suspicious and confusing for new users.

taavirehemagi:
Should "ask for project also be disabled" ?

taavirehemagi:
Closed in #1096

Issue #1087, opened 2015-12-09
Edit screen should be opened automatically when timer is started


kyllikekimmel:
Otherwise it's difficult for new users to figure out where to add a description and project to the time entry.

alfonsogarciacaro:
Related to #1040

taavirehemagi:
Closed in #1096

Issue #1074, opened 2015-12-04
Description is overlapping duration on the toolbar.


taavirehemagi:
<img width="355" alt="screen shot 2015-12-04 at 14 07 03" src="https://cloud.githubusercontent.com/assets/2154171/11589333/530f052c-9a90-11e5-9576-ee454f37708e.png">

taavirehemagi:
Closed in #1096

PR #1096, opened 2015-12-14
Small fixes and improvements

### LINKED ISSUE #1094: Remove "Alpha state" from app settings
This is confusing to many of our users (including me). Thank you.

taavirehemagi:
Closes #1094, #1074, #952, #1087, #1095, #1088, #1097


taavirehemagi:
Default state of settings for new user in this PR.
screen shot 2015-12-14 at 15 55 39

Joey/UI/Activities/EditTimeEntryActivity.cs
```diff
@@ -59,6 +62,12 @@ protected override void OnCreateActivity (Bundle state)
                     .Commit ();
                 }
             }
+
+            if (openProjects) {
```
monday8am: you can check Settings directly, you don't need to pass from the previous activity (IMO)
`var settingsStore = ServiceContainer.Resolve<SettingsStore> ();`

Joey/UI/Fragments/LogTimeEntriesListFragment.cs
```diff
@@ -109,6 +110,18 @@ public override View OnCreateView (LayoutInflater inflater, ViewGroup container,
             if (ViewModel.HasMore) {
                 OBMExperimentManager.Send (OBMExperimentManager.HomeEmptyState, "startButton", "click");
             }
+
+            if (ViewModel.IsTimeEntryRunning) {
+
+                var intent = new Intent (Activity, typeof (EditTimeEntryActivity));
+                IList<string> guids = ((TimeEntryHolder)logAdapter.GetEntry (1)).TimeEntryGuids;
+                intent.PutStringArrayListExtra (EditTimeEntryActivity.ExtraGroupedTimeEntriesGuids, guids);
```
monday8am: in this method you have this line:
`await ViewModel.StartStopTimeEntry ();`

you could replace it by:
`var timeEntry = await ViewModel.StartStopTimeEntry ();`

you can use the timeEntry.Id to pass it to the edit class (more reliable than the `logAdapter`). And you always *start* single time entries, not groups, group checking maybe is not needed.

Joey/UI/Activities/EditTimeEntryActivity.cs
```diff
@@ -59,6 +62,12 @@ protected override void OnCreateActivity (Bundle state)
...
+                var intent = new Intent (this, typeof (ProjectListActivity));
+                intent.PutExtra (BaseActivity.IntentWorkspaceIdArgument,  guids[0]);
+                StartActivityForResult (intent, 0);
+            }
         }
```
monday8am: This solution fails because the data is saved in the intent and it tries to open the ProjectListActivity again and again. In a future we'll use fragments :) .. for the moment, we should use a *one-time* flag or something similar.

taavirehemagi:
I think it's better now. What  do you think?

monday8am:
I'll take a look right now to merge it. Just rebase it with master (a terrible task :dancer:) ..

taavirehemagi:
The change works good for me! :+1: Can we merge? @alfonsogarciacaro @monday8am

Issue #1107, opened 2015-12-16
Add more tests for single entries TimeEntriesCollectionView


alfonsogarciacaro:
Make sure that all situations are covered. Problems are coming mainly from Move events so combine moves (both forward and backward) with other operations like adding and removing.

alfonsogarciacaro:
Closed by [this PR](https://github.com/toggl/mobile/pull/1111)

PR #1111, opened 2015-12-16
Add more tests for TimeEntriesCollection class

alfonsogarciacaro:
This closes issue: Add more tests for single entries TimeEntriesCollectionView #1107
It also includes some fixes to the Diff algorithm. Mainly it gives preference to date headers to remain static when calculating the diffs.


monday8am:
@alfonsogarciacaro  more tests and a bit more complex algorithm. In theory ok but I'm totally sure that we'll improve this code a bit more ;)

PR #1129, opened 2015-12-28
New offline mode approach. Optimised LoadMore method.

monday8am:
Currently the operations to download data from remote could be described in this way: 
- A *Date pagination* is used to get data from remote: A date _A_ is defined and the app receive entries for the las 5 days. 
- If the request fails, a fixed amount of entries is taken from local.
- An extra call to get is needed if the request fails and the amount of time entries presented is bigger than the period (Date - 5 days) requested to server.

With the PR, the new steps are:
- A *Date pagination* is used to get data from local and presented.
- At the same time, the same period of time is requested to the server in the background.
- When the request is ok, the loca *Date pagination* is adjusted. 

Advantages: In the first case, you have to wait for a remote request to show some data, in the second case, data is taken from local and updated if the request is ok.


monday8am:
no comments, I guess you tested and everything was ok :+1: :dancer:

PR #1137, opened 2015-12-30
Refactor of all classes related with Project list.

monday8am:
Complete refactor of `WorkspaceProjectsView.cs`. I have created it again from scratch, this time based on `ObservableRangeCollection`, deleting composed objects and only  extending the existing `CommonData` class. Everything is functional except the *Most used projects* section. 
Highlights:
- Search uses Rx, please @alfonsogarciacaro take a look to it:
https://github.com/toggl/mobile/blob/anton-projectlist/Phoebe/Data/ViewModels/ProjectListViewModel.cs#L46
- I have removed the update feature (the screen could update if the server info is updated) and maybe it will not be restored. In case of projects, is not a common behaviour. 
- Now, because the `ProjectCollection` is based in an `ObservableCollection`, the tasks are animated correctly. This will be helpful in iOS too.

@taavirehemagi probably you'll find some bugs. Just let me know to fix it.


alfonsogarciacaro:
I don't see any problem with the use of Rx in this context. Maybe a `Subject` could be used instead of a private event but this shouldn't be much of a problem. Also, the stream is not being scheduled, which means some events could race against each other, but that won't happen now because [there's one single caller of `SearchByProjectName`](https://github.com/toggl/mobile/blob/d4b17ffd4142642c133451bd9957924d51e02c47/Joey/UI/Fragments/ProjectListFragment.cs#L210).
One problem may be that if `Init` hasn't finished before [`onSearch` is called](https://github.com/toggl/mobile/blob/master/Phoebe/Data/ViewModels/ProjectListViewModel.cs#L77), the delegate may be null (as nobody has subscribed yet) so [a null check would be necessary](http://stackoverflow.com/a/672661/3922220).

Issue #1126, opened 2015-12-23
Add Railgun bug tracker in replacement of Xamarin Insights


monday8am:
Closes #1064

monday8am:
@alfonsogarciacaro take care of it:
- Remove Xamarin libs, add RayGun nuget packages and check if is possible to work with it from the common project.
- Modify the `LogClient`  (https://github.com/toggl/mobile/blob/master/Phoebe/Logging/LogClient.cs) to work with RayGun.
- Configure the Build.cs with the RayGun keys and let me know to update Teamcity.

alfonsogarciacaro:
I've created [a branch](https://github.com/toggl/mobile/compare/alfonso-raygun?expand=1) to use Raygun. Things to note:
- Not tested yet on device.
- Nuget tools in Xamarin didn't set the correct references, so I edited the .csproj files manually. Probably this will be needed again when creating the package.
- Apparently, Raygun doesn't have two different methods (Track and Report) to send messages, so [I've used tags](https://github.com/toggl/mobile/blob/alfonso-raygun/Phoebe/Logging/LogClient.cs#L47-L48) to make the distinction regarding severity.
- For the Desktop project, I've added a compiler directive to ignore the file, so there's no need to make a reference to Raygun. Another solution would be to register the LogClient from Joey and Ross.

monday8am:
closed by #1138

PR #1138, opened 2015-12-31
Replace Xamarin Insights with Raygun API

alfonsogarciacaro:
- Tested in Android, not iOS (should work)
- Nuget tools in Xamarin didn't set the correct references, so I edited the .csproj files manually. Probably this will be needed again when creating the package.
- Apparently, Raygun doesn't have two different methods (Track and Report) to send messages, so I've used tags to make the distinction regarding severity.
- For the Desktop project, I've added a compiler directive to ignore the file, so there's no need to make a reference to Raygun. Another solution would be to register the LogClient from Joey and Ross.


alfonsogarciacaro:
Closes #1126

PR #1141, opened 2016-01-03
Android wear app signed correctly in Teamcity

monday8am:
The Android wear app (Chandler) was not installed because the wearable-app.apk (inside res/raw/) was not correctly signed. I have changed the process of *build* to:
- build an unsigned .apk including the wearable apk (Chandler project) and all conf files needed.
- unpack the unsigned apk
- sign the wearable apk correctly.
- pack again and sign the container apk (Joey project) and align the zip.

Seems to be Xamarin Studio does the process in this way. Now we need to test that the app is installed correctly in wear device.


monday8am:
@taavirehemagi please, generate an .apk of this branch using Teamcity and let me know if it works

PR #1187, opened 2016-01-19
Double check every item if it's synced or not.

taavirehemagi:
Checks RemoteId value in case of new entries and IsDirty for old entries.
Maybe UI needs some work. Suggestions?
screen shot 2016-01-19 at 15 56 34


taavirehemagi:
@pe0ny any quick suggestions?

taavirehemagi:
bump

PR #1200, opened 2016-01-26
Model classes modified. Mapper finished.

monday8am:


monday8am:
@alfonsogarciacaro Maybe you need extra job to merge the branch, the important points are:
- `TogglRestClient` and `ITogglRestClient` now work with `CommonData` and not with `CommonJson`. The conversion is done inside, in the generic methods like https://github.com/toggl/mobile/blob/anton-json/Phoebe/_Net/TogglRestClient.cs#L256-L274 . This functionality is not tested.

- The Mapper class is working ok in both ways: Data -> Json  and Json -> Data. Maybe I forgot some field but the basic tests worked ok. You can use this class to map in both ways in the Store when JSON objects come or when you want to send this objects to server, in case you want to use the old `TogglRestClient` (it would need some modifications anyway)

- There is a new ID duplication like *WorkspaceRemoteId* and *WorkspaceId*. You can detect the direction of the messages according to which ID is null or not (detect between Update and Create).

Issue #1220, opened 2016-02-05
Description change gets reverted when editing an entry


alfonsogarciacaro:
How to reproduce:
- Select an entry and enter edit screen
- Edit the description
- Now select either the start and stop time and change it, touch DONE
- When the timer picker is dismissed, notice the description has gone back to its previous state

monday8am:
Check if the bug appears when a *new* time entry is added.

kyllikekimmel:
Steps to reproduce on a new time entry: 
* No-user mode, Grouped entries selected
* Start a timer, add a description and an existing project
* Tap Back arrow
* Time entry looks as it should with the description and project attached 
* When that entry is stopped (from the Stop button), the project and description disappear, leaving the entry with (no project) and (no description).

PR #1268, opened 2016-02-15
Save description after editing.

taavirehemagi:
When you edit description, and then start to edit Start or stop time, the Description field in EditTimeEntryViewModel gets overwritten in https://github.com/toggl/mobile/blob/master/Phoebe/Data/ViewModels/EditTimeEntryViewModel.cs#L139
(temporary solution is to save the Description at the moment of editing (like all other fields).
Is there a cleaner way?

Closes https://github.com/toggl/mobile/issues/1220

### DIFF
--- a/Joey/UI/Fragments/EditTimeEntryFragment.cs
+++ b/Joey/UI/Fragments/EditTimeEntryFragment.cs
@@ -129,7 +129,11 @@ namespace Toggl.Joey.UI.Fragments
             // the event "CheckedChange" isn't found when
             // the app is compiled for release. Investigate!
             BillableCheckBox.CheckedChange += (sender, e) => {
-                ViewModel.IsBillable = BillableCheckBox.Checked;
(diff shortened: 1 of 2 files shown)


monday8am:
Description is overwritten here, every time another property is changed:
https://github.com/toggl/mobile/blob/master/Phoebe/Data/ViewModels/EditTimeEntryViewModel.cs#L216

The correct way is to make `Description` and `Billable` as the other properties, only one way binding and modify them calling explicit methods. 

The steps:
- Remove double way binding for both properties.
- Add two methods to modify them, like this methods:
https://github.com/toggl/mobile/blob/master/Phoebe/Data/ViewModels/EditTimeEntryViewModel.cs#L130-L169
- Remove these two lines:
https://github.com/toggl/mobile/blob/master/Phoebe/Data/ViewModels/EditTimeEntryViewModel.cs#L183-L184
- Call the method to modify description every time the content of the textfield changes.

taavirehemagi:
ok, will modify the code :+1:

PR #1290, opened 2016-02-17
No-user mode implementation

taavirehemagi:


Joey/UI/Activities/EditTimeEntryActivity.cs
```diff
@@ -61,5 +64,14 @@ protected override void OnCreateActivity (Bundle state)
...
             }
         }
+
+        public void ShowSoftKeyboard (View input, bool selectText)
+        {
+            if (selectText) { ((EditText)input).SelectAll(); }
+            ThreadPool.QueueUserWorkItem (s => {
+                Thread.Sleep (100); // For some reason, a short delay is required here.
+                RunOnUiThread (() => ((InputMethodManager)GetSystemService (InputMethodService)).ShowSoftInput (input, ShowFlags.Implicit));
+            });
```
alfonsogarciacaro: Do you just need the 100 ms delay here? If so, as you're already on UI Thread, this should suffice:
```
await System.Threading.Tasks.Task.Delay (100);
((InputMethodManager)GetSystemService (InputMethodService))
    .ShowSoftInput (input, ShowFlags.Implicit);
```

Joey/UI/Activities/MainDrawerActivity.cs
```diff
@@ -249,6 +260,13 @@ private void OpenPage (int id)
             }
         }
 
+        public void ForgetCurrentUser ()
+        {
+            var authManager = ServiceContainer.Resolve<AuthManager> ();
+            authManager.Forget ();
+            StartAuthActivity ();
+        }
```
alfonsogarciacaro: Already talked with Anton, but it's better not to use `AuthManager` as a way to keep state. We're having problems with the order in which the services are getting ready. Also there's a bit of confusion in this class, in L85 `authManager` is cached but here is resolved explicitly.

Joey/UI/Adapters/DrawerListAdapter.cs
```diff
@@ -108,6 +121,28 @@ public override int GetItemViewType (int position)
...
+                    foreach (var sub in item.SubItems) {
+                        if (sub.VMode == VisibilityMode.Normal && authManager.OfflineMode || sub.VMode == VisibilityMode.Offline && !authManager.OfflineMode) {
+                            continue;
+                        }
+                        subItems.Add (sub);
+                    }
+                    item.SubItems = subItems.Count > 0 ? subItems : null;
+                }
+            }
+            return newList;
```
alfonsogarciacaro: Suggestion: use LINQ here.
```
            Func<DrawerItem, bool> filter = item =>
                !(item.VMode == VisibilityMode.Normal && authManager.OfflineMode) &&
                !(item.VMode == VisibilityMode.Offline && !authManager.OfflineMode);

            return list.Where (filter)
                       .Select (item => item.With (item.SubItems.Where (filter).ToList ()))
                       .ToList ();
```

This method must be added to DrawerItem (it can be modified to add more optional arguments):
```
            public DrawerItem With (List<DrawerItem> subItems)
            {
                return new DrawerItem {
                            Id = this.Id,
                            TextResId = this.TextResId,
                            ImageResId = this.ImageResId,
                            ChildOf = this.ChildOf,
                            IsEnabled = this.IsEnabled,
                            Expanded = this.Expanded,
                            VMode = this.VMode,
                            SubItems = subItems.Count > 0 ? subItems : null
                };
            }
```

Joey/UI/Fragments/RegisterUserFragment.cs
```diff
@@ -0,0 +1,602 @@
...
+
+        private void SyncRegisterButton ()
+        {
+            var enabled = !isAuthenticating;
+
+            if (String.IsNullOrWhiteSpace (EmailEditText.Text) || !EmailEditText.Text.Contains ('@')) {
+                enabled = false;
+            } else if (String.IsNullOrWhiteSpace (PasswordEditText.Text) || PasswordEditText.Text.Length < 6) {
+                enabled = false;
+            }
```
alfonsogarciacaro: We may ask the backend guys if they have some regular expressions to validate the email and password and use them here.

refiito: e-mail validation with a regular expression isn't that...

well, basically, here's what we're using:
`^[^<>\\#$@\s]+@[^<>\\#$@\s]*[^<>\\#$\.\s@]{1}?\.{1}?[^<>\\#$\.@\s]{1}?[^<>\\#$@\s]+$`

alfonsogarciacaro: Thanks, Margus! Then the check would become (I'd put the regex pattern somewhere else though):
```
var emailValidator = "^[^<>\\\\#$@\\s]+@[^<>\\\\#$@\\s]*[^<>\\\\#$\\.\\s@]{1}?\\.{1}?[^<>\\\\#$\\.@\\s]{1}?[^<>\\\\#$@\\s]+$";

RegisterButton.Enabled =
      isAuthenticating &&
      Regex.IsMatch (EmailEditText.Text ?? "", emailValidator) &&
      (PasswordEditText.Text ?? "").Length >= 6;
```

Phoebe/Net/SyncManager.cs
```diff
@@ -88,6 +94,42 @@ private void OnAuthChanged (AuthChangedMessage msg)
...
+            // Unsubscribe from models commited messages (our actions trigger them as well,
+            // so need to ignore them to prevent infinite recursion.
+            if (subscriptionDataChange != null) {
+                bus.Unsubscribe (subscriptionDataChange);
+                subscriptionDataChange = null;
+            }
+
+            try {
+                // Make sure that the RunInBackground is actually started on a background thread
+                await await Task.Factory.StartNew (() => MergeNoUserData ());
```
alfonsogarciacaro: Duplicated `await`

Phoebe/Net/SyncManager.cs
```diff
@@ -147,6 +189,72 @@ private void OnAuthChanged (AuthChangedMessage msg)
...
+                    String.Concat ("UPDATE ", pudTbl, " SET UserId = ?")
+                    ,user.Id);
+
+
+                // Delete old data.
+                var usrTbl = await store.GetTableNameAsync<UserData>();
+                var wsTbl = await store.GetTableNameAsync<WorkspaceData>();
+                await store.QueryAsync<WorkspaceData> (String.Concat ("DELETE FROM ", wsTbl ," WHERE Name='Workspace'"), workspace.Result.Id );
+                await store.QueryAsync<WorkspaceData> (String.Concat ("DELETE FROM ", usrTbl ," WHERE Name='offlineUser'"), workspace.Result.Id );
+            } catch (Exception e) {
```
alfonsogarciacaro: It's better (more performant and safer to prevent intermediate states if there's a fail) if non-select queries are executed in a transaction. I'd replace L207-L241 with:
```
await store.ExecuteInTransactionAsync (ctx => {
                    var user = serverData.User.Import (ctx);
                    var workspace = serverData.Workspaces.ElementAt (0).Import (ctx);

                    var workspaceId = Tuple.Create ("WorkspaceId", (object)workspace.Id);
                    var userId = Tuple.Create ("UserId", (object)user.Id);

					//Lift data from the local objects to server objects
                    updateTable<ProjectData> (ctx, workspaceId);
                    updateTable<ClientData> (ctx, workspaceId);
                    updateTable<TagData> (ctx, workspaceId);
                    updateTable<TimeEntryData> (ctx, workspaceId, userId);
                    updateTable<ProjectUserData> (ctx, userId);

                    // Delete old data.
                    deleteTable<UserData> (ctx, Tuple.Create ("Name", (object)"Workspace"));
                    deleteTable<WorkspaceData> (ctx, Tuple.Create ("Name", (object)"offlineUser"));
                });
```

You'll need the following helper methods:
```
        private int updateTable<T> (IDataStoreContext ctx, params Tuple<string, object>[] args)
        {
            var query = string.Format (
                "UPDATE {0} SET {1}",
                ctx.Connection.GetMapping<T> ().TableName,
                string.Join (",", args.Select (x => x.Item1 + "=?"))
            );
            return ctx.Connection
                      .CreateCommand (query, args.Select (x => x.Item2).ToArray ())
                      .ExecuteNonQuery ();
        }

        private int deleteTable<T> (IDataStoreContext ctx, params Tuple<string, object>[] conditions)
        {
            var query = string.Format (
                "DELETE FROM {0} WHERE {1}",
                ctx.Connection.GetMapping<T> ().TableName,
                string.Join (" AND ", conditions.Select (x => x.Item1 + "=?"))
            );
            return ctx.Connection
                      .CreateCommand (query, conditions.Select (x => x.Item2).ToArray ())
                      .ExecuteNonQuery ();
        }
```

taavirehemagi:
This is rebased to the current master state. Ready to merge?

PR #1315, opened 2016-02-26
Empty workspace patch.

monday8am:
Temporal patch (only 8.1.2 - 7.1.0 release) to repair empty workspaces and to avoid them.

### DIFF
--- a/Phoebe/Data/SqliteDataStore.cs
+++ b/Phoebe/Data/SqliteDataStore.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using SQLite.Net;
 using SQLite.Net.Async;
 using SQLite.Net.Interop;
+using Toggl.Phoebe.Data.DataObjects;
 using Toggl.Phoebe.Net;
 using XPlatUtils;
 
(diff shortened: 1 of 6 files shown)


Phoebe/Data/SqliteDataStore.cs
```diff
@@ -47,12 +48,30 @@ private void CreateTables ()
...
+            if (user != null && user.DefaultWorkspaceId != Guid.Empty) {
+                var tableNames = new List<string>  { cnn.Table<TimeEntryData>().Table.TableName,
+                                                     cnn.Table<ClientData>().Table.TableName,
+                                                     cnn.Table<ProjectData>().Table.TableName,
+                                                     cnn.Table<TagData>().Table.TableName,
+                                                     cnn.Table<TaskData>().Table.TableName
+                                                   };
+                cnn.RunInTransaction (() =>  {
+                    foreach (var tableName in tableNames) {
+                        var q = string.Concat ("UPDATE ", tableName ," SET WorkspaceId = '", user.DefaultWorkspaceId ,"' WHERE WorkspaceId = '", Guid.Empty, "'");
```
alfonsogarciacaro: It's a bit safer (an probably more performant) to work with parameters instead of building the whole string:
```
cnn.Execute ($"UPDATE {tableName} SET WorkspaceId=? WHERE WorkspaceId=?", user.DefaultWorkspaceId, Guid.Empty);
```
> Note the table name cannot be passed as a parameter

Tests/Data/DataStoreTest.cs
```diff
@@ -198,5 +199,179 @@ public void TestTransactionRollback ()
...
+            if (user != null && user.DefaultWorkspaceId != Guid.Empty) {
+                var tableNames = new List<string>  { cnn.Table<TimeEntryData>().Table.TableName,
+                                                     cnn.Table<ClientData>().Table.TableName,
+                                                     cnn.Table<ProjectData>().Table.TableName,
+                                                     cnn.Table<TagData>().Table.TableName,
+                                                     cnn.Table<TaskData>().Table.TableName
+                                                   };
+                cnn.RunInTransaction (() =>  {
+                    foreach (var tableName in tableNames) {
+                        var q = string.Concat ("UPDATE ", tableName ," SET WorkspaceId = '", user.DefaultWorkspaceId ,"' WHERE WorkspaceId = '", Guid.Empty, "'");
```
alfonsogarciacaro: Same as above

PR #1388, opened 2016-03-17
added script to replace secret build variables, closes #1381

### LINKED ISSUE #1381: Create script to insert secret API keys into cs config file

amulware:
The script includes a map of secrets to replace and will fail if not all secrets are defined as environmental variables.

Let me know if I should add more secrets before this is merged.


amulware:
By now, the script replaces all the secret variables (I know about) for both android and ios (there is a script for each). As far as I can tell, this is done.

PR #1411, opened 2016-03-28
Implement fragment-only navigation. Use transitions to move from one screen to another.

taavirehemagi:
- TE list click - expands the list item to an open edit fragment [Gmail imitation] (at some iteration I was sharing elements (duration and description), but at the end I removed them because it looks a lot cleaner without them). Closes 
- Manual button click fades the edit screen
- FAB in Timer and Project list screen uses SlideBottom (seems to be common for FAB animations in a lot of apps)
- Edit screen project item -> project list uses fade. 
- All return / exit's use fade


monday8am:
Out of date. I'll close the PR and keep the branch (maybe it is needed later)

PR #1428, opened 2016-04-04
Restore spining loader to RecyclerView.

taavirehemagi:


monday8am:
@taavirehemagi we need to check the authManager to show an spinner or not :( .. we'll change that soon .

PR #1486, opened 2016-04-15
implemented rudimentary settings migration

amulware:
This may not be done yet, but it should mostly work. Also wants more testing I imagine.


Phoebe/Helpers/ServiceContainer.cs
```diff
@@ -70,6 +70,38 @@ public static void RegisterScoped<T> (T service)
             services[typeof(T)] = service;
         }
 
+        public static bool TryResolve<T>(out T t)
+        {
+            object o;
+            var result = TryResolve(typeof(T), out o);
+            t = (T)o;
```
alfonsogarciacaro: To avoid casting null you could use: `t = result ? (T)o : null;` (in this case a `T : class` constraint would be needed, which is indeed desirable in this case)

alfonsogarciacaro:
I find it a bit confusing that `SettingsStore` implements `IOldSettingsStore` in Android but `OldSettingsStore` is used in the case of iOS, is this intended?

Besides that, although not strictly necessary it would be ideal to have some kind of symmetry with the db migration. Right now they seem to be very different process. We could also use some kind of versioning for the settings too like we did for the db, but if it's ok to reset the settings in the case of conflict (e.g., we change again the settings in the future but a user has skipped the version in between) it shouldn't be that important.

Phoebe/Helpers/ServiceContainer.cs
amulware: Can you elaborate? Casting `null` to any type works just fine. I don't see why the class constraint is desirable. It may be that there are no value type things to resolve, but does it matter?

The only way in which the cast could fail is if it tries performing an invalid cast, which means that someone somehow put an object of the wrong type into the dictionary, which should not be possible (and I think SHOULD cause a failure).

amulware:
@monday8am any reason for the difference in names between the two classes?

Regarding the symmetry with the db migration, I'm not sure that could work. The two seem like very different processes that only 'happen' to be both part of this new build. In the future we may have new database versions for which we will add migration logic, but the way settings are stored should not change anytime soon, should it? Adding more settings properties should be no issue and not required any migration. They would simple be filled with default values.

Phoebe/Helpers/ServiceContainer.cs
alfonsogarciacaro: Null casting shouldn't cause any error but I think the change can make the intention of `TryResolve` clearer: that is, we have to deal with `null` if `result` is false. The `class` constraint is desirable because `TryResolve(Type type, out object service)` will set `service` to `null` if the service is not found and in this case a casting to a value type will do fail. That's why it's safer to restrict `TryResolve<T>` to nullable types.

amulware: Should we add the same constraint to the `Resolve` methods then?

alfonsogarciacaro: The `Resolve(Type type)` method doesn't return `null` but [throws an exception](https://github.com/toggl/mobile/blob/master/Phoebe/ServiceContainer.cs#L101) when the service is not found so I guess it's not needed there.

monday8am:
superseded by #1500

PR #1500, opened 2016-04-26
Settings migration using Paul's #1486 as a base.

monday8am:


Phoebe/Data/SettingChangedMessage.cs
```diff
@@ -1,19 +0,0 @@
-﻿using System;
-
-namespace Toggl.Phoebe.Data
-{
-    public class SettingChangedMessage : Message
```
alfonsogarciacaro: @monday8am, why has this been deleted? It's needed for Android to compile.

alfonsogarciacaro: @anton, I just noticed `SettingChangedMessage` is only used for the `MessageBus`. Are we removing completely the `MessageBus` in `unidirectional`? It seems to be used in the Migration tests.

PR #1501, opened 2016-04-27
Double.TryParse bug. Fixed with InvariantCulture parameter.

jespersh:
…double.TryParse as cultures can use comma for decimal seperator, like da-DK.

See the repository https://github.com/jespersh/test-toggle-float for an example bug.


monday8am:
Hi @jespersh, sorry my delay, it is a fantastic PR because it fixes a bug that was breaking us the head. I'll merge it right now. Thanks again!

PR #1520, opened 2016-05-03
Added sync status view and better error handling

monday8am:


Phoebe/Reactive/SyncManager.cs
```diff
@@ -84,11 +84,16 @@ public static void Cleanup()
...
-                .Subscribe(_ => { }, ex => logError(ex));
+            StoreManager.Singleton
+            .Observe()
+            .SelectAsync(EnqueueOrSend)
+            .Subscribe((_) => { },
+            (ex) =>
+            {
+                logError(ex, "Failed to sync. Main observer bug.");
+                // Should pass a generic request?
+                RxChain.Send(new DataMsg.ServerResponse(new ServerRequest.GetChanges(), ex));
```
alfonsogarciacaro: Not sure it's necessary to pass this message to `RxChain`. This point will only be reached if an uncaught exception leaks from `EnqueOrSend` and there's no `ServerRequest.GetChanges` message sent from there. Is this to notify views expecting a response?

monday8am: yes, if not, the view keeps the spinner visible. We should at least hide the spinner and notify that there was an error.

alfonsogarciacaro: It's ok to leave this here. I would try to improve the error message (`$"Unexpected error in {nameof(SyncManager)}"`) and remove or make the comment more detailed (`// Send a generic response in case some views (e.g. a loading spinner) are waiting for it`). But we'll probably have to wrap [this lines](https://github.com/toggl/mobile/blob/unidirectional/Phoebe/Reactive/SyncManager.cs#L161-L188) in a `try...catch`.

Phoebe/Reactive/SyncManager.cs
```diff
@@ -115,7 +120,13 @@ public static void Cleanup()
                         throw new Exception("Unexpected Authenticate operation");
                 }
             }))
-            .Subscribe(_ => { }, ex => logError(ex));
+            .Subscribe((_) => { },
+            (ex) =>
+            {
+                logError(ex, "Failed to sync. RequestManager bug.");
+                // Should pass a dummy request for this cases?
+                RxChain.Send(new DataMsg.ServerResponse(new ServerRequest.GetChanges(), ex));
```
alfonsogarciacaro: No harm to have this here but all methods handling requests (`DownloadEntries`, `GetChanges`, `AuthenticateAsync`) are wrapped by `try...catch` this point should never be hit. Se comments above for suggestions to improve error message and comment.

PR #1525, opened 2016-05-04
Implement design for Migration screen. Small bugs fixed.

monday8am:


Phoebe/Reactive/Reducers.cs
```diff
@@ -506,9 +507,40 @@ static DataSyncMsg<AppState> InitStateAfterMigration(AppState state, DataMsg msg
             dataStore.Table<TaskData>().ForEach(x => tasks.Add(x.Id, x));
             dataStore.Table<TagData>().ForEach(x => tags.Add(x.Id, x));
 
+            var settings = state.Settings;
```
monday8am: @alfonso @paul Now old settings are loaded after the state is fulfilled with the old data. In that way, if we do it before it generates contradictions like `Settings.UserId != null` but no user is present in DB.

Phoebe/Data/DatabaseHelper.cs
```diff
@@ -151,13 +167,10 @@ public static int CheckOldDb(string dbDir)
...
                 }
             }
-            catch (MigrationException ex)
-            {
-                resolveMigrateException(ex, newDB, dbDir);
-            }
             catch (Exception ex)
             {
-                var ex2 = new MigrationException("Unknown exception during migration", ex);
+                var ex2 = (ex is MigrationException) ? (MigrationException)ex :
```
monday8am: @alfonso the double catch was not working correctly.
https://insights.xamarin.com/app/Toggl-iOS/issues/1488

Phoebe/Data/DatabaseHelper.cs
```diff
@@ -25,10 +25,25 @@ public static string GetDatabaseDirectory()
...
+            for (int i = 0; i < version; i++)
+            {
+                // Delete previous version if exist.
+                var dir = GetDatabaseDirectory();
+                var path = GetDatabasePath(dir, i);
+                if (FileExists(path))
+                    File.Delete(path);
+            }
+        }
+
```
monday8am: @paul @alfonso After the user decides to discard not synced data and get a fresh copy from server, this method is called to deleted all previous DB versions.

PR #1536, opened 2016-05-10
Anton noapi

monday8am:
@taavirehemagi only the last two commits are mine, maybe you should rebase unidirectional-nouser branch.


taavirehemagi:
This is how the drawer should look like in no-user mode. No need to confuse users with dummy data I think
https://app.zeplin.io/project.html#pid=56a8c0ec4889aa4455dae843&sid=56e93260c84473d610b86c57

PR #1544, opened 2016-05-13
Migration detection and migration fragment added.

monday8am:


Joey/UI/Activities/MainDrawerActivity.cs
```diff
@@ -186,6 +165,56 @@ public override void OnBackPressed()
...
+        }
+
+        private bool tryMigrateDatabase(IUserData userData)
+        {
+            var oldVersion = DatabaseHelper.CheckOldDb(DatabaseHelper.GetDatabaseDirectory());
+            if (oldVersion == -1)
+                return false;
+
+            var migrationFragment = MigrationFragment.Init(oldVersion, SyncSqliteDataStore.DB_VERSION);
+            OpenFragment(migrationFragment);
```
monday8am: @alfonsogarciacaro I have added the migration detection methods and a dummy migration fragment. You can use it if you want to implement the migration screen.

Joey/UI/Fragments/FeedbackNoApiFragment.cs
```diff
@@ -12,6 +12,14 @@ public class FeedbackNoApiFragment : Fragment
...
         private Button noUserRegisterButton;
 
+        public FeedbackNoApiFragment()
+        {
+        }
+
+        public FeedbackNoApiFragment(IntPtr jref, Android.Runtime.JniHandleOwnership xfer) : base(jref, xfer)
+        {
+        }
+
```
taavirehemagi: Are these constructors necessary? they are not present any more in LogTimeEntryListFragment for example...

Joey/UI/Adapters/DrawerListAdapter.cs
```diff
@@ -191,7 +183,7 @@ protected override void Rebind()
         public enum VisibilityMode
         {
             Normal,
-            Offline,
+            NoApiToken,
```
taavirehemagi: In this case the wording "noApiToken" seems to specific and doesn't capture the meaning of the state. Maybe Connected and NotConnected over "Normal" and "Offline". Also, what if in the future we will use a different way of determining if a user is connected or not?

Joey/UI/Fragments/FeedbackNoApiFragment.cs
monday8am: just check this commit from 15 days ago:
https://github.com/toggl/mobile/commit/932696f3511baae984d3b18ebb396f6f0e917341

or the same file in master:
https://github.com/toggl/mobile/blob/master/Joey/UI/Fragments/LogTimeEntriesListFragment.cs

Or this search in XI:
https://insights.xamarin.com/app/Toggl-Android/issues?s=Unable%20to%20activate%20instance%20of%20type%20

taavirehemagi: alright. if it's still necessary then good with me 👍  I just didn't see any examples in unidirectional mode anymore and figured that the need for this is over. fine with me then

PR #1604, opened 2016-05-24
added manual mode to LogTimeEntriesVM and iOS, fixes #1598

### LINKED ISSUE #1598: iOS 7.1.0: Adding entries manually
It's not possible to manually add a new entry in v7.1.0. This feature must be brought back.

Some users also requested that it's possible to set the start day when adding the entry (this also affects entry edition).

amulware:
method in view model can be used by android as well if appropriate

### DIFF
--- a/Phoebe/ViewModels/LogTimeEntriesVM.cs
+++ b/Phoebe/ViewModels/LogTimeEntriesVM.cs
@@ -13,6 +13,7 @@ using Toggl.Phoebe.ViewModels.Timer;
 using XPlatUtils;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Collections.Generic;
 
 namespace Toggl.Phoebe.ViewModels
 {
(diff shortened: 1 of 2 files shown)


Phoebe/ViewModels/LogTimeEntriesVM.cs
```diff
@@ -168,6 +169,30 @@ public void RemoveTimeEntry(int index)
...
+        public Task<ITimeEntryData> ManuallyCreateTimeEntry()
+        {
+            // TODO: should this be in a reducer instead?
+            var state = StoreManager.Singleton.AppState;
+            var draft = state.GetTimeEntryDraft();
+            var now = Time.UtcNow.Truncate(TimeSpan.TicksPerSecond);
+            var newEntry = TimeEntryData.Create(draft: draft, transform: x =>
+            {
+                x.State = TimeEntryState.Finished;
+                x.StartTime = now;
```
alfonsogarciacaro: Shall we make the StartTime 5 minutes earlier by default or something like that?

amulware: I suppose we could, though it is somewhat arbitrary, and how often will someone actually want to track only 5 minutes? The other apps also do not make any such assumptions. Maybe this is something to have a setting for in the future.

PR #1615, opened 2016-05-26
Restore autocomplete functionality 

taavirehemagi:


Phoebe/ViewModels/EditTimeEntryVM.cs
```diff
@@ -116,6 +136,33 @@ public string ProjectColorHex
...
+        }
+
+        public List<TimeEntryData> LoadSuggestions(string description)
+        {
+            var suggestions = new List<TimeEntryData> ();
+            if (description.Length < 3)
+                return suggestions;
+
+            foreach (var s in allSuggestions)
+                if (s.Description.Contains(description))
```
amulware: I propose changing this line to:
`if (s.Description.IndexOf(description, StringComparison.OrdinalIgnoreCase) != -1)`
to make the check case-insensitive.

PR #1617, opened 2016-05-27
fix empty state layout, closes #1589

### LINKED ISSUE #1589: iOS onboarding experiment is not displayed properly
It seems that the design is displayed slightly to the right. Screenshots taken on iPhone 5c.

The arrow is missing in the "included" group view: 
image

The "excluded" group view (default view) also seems to be displayed slightly to the right: 
image

amulware:


amulware:
oops. wrong base branch.

PR #1623, opened 2016-05-30
Add throttle to Autocomplete

alfonsogarciacaro:
Last week I wanted to add a Throttle to Autocomplete, but it wasn’t working because I forgot `ObserveOn`. This prevents too many searches in the database if the user types too fast.


monday8am:
@alfonsogarciacaro I have a couple of suggestions. I'll send them after the meeting.

PR #1624, opened 2016-05-30
Fix #1594

### LINKED ISSUE #1594: iOS 7.1.0: "Sliding Drawer" gesture conflicting with "Continue Entry" swipe
A user mentioned that many times he tried to continue an entry (by swiping it to the right) the sliding drawer was open instead. He's suggesting to disable the gesture or making the recognition area smaller. The issue can't be reproduced in the simulator

> the sliding drawer that reveals the settings by sliding from the left. Disable this or reduce the gesture recognition area greatly. When modifying start/end times and swiping up &amp; down, the drawer keeps trying to reveal the settings menu. If the settings menu was something I used very often, this might be understandable but I&#39;ve opened the setting only two or three times in the thousands of times I&#39;ve opened and used the app. This is not a useful feature, I can tap the top left icon to reveal the settings if I need to; I don&#39;t need it accessible from any point on the screen.

Case 290782

alfonsogarciacaro:
Naive attempt to fix #1594 by disabling left pan gesture on `MainViewController`. Please review if this is the best way to do it and whether we should completely remove the code for it, including `MainViewController.OnLeftPanGesture`.


monday8am:
@alfonsogarciacaro if you tested it and it works, for me we can close it.

taavirehemagi:
@alfonsogarciacaro 
the method that is responsible for dragging the drawer open is here:
https://github.com/toggl/mobile/blob/unidirectional/Ross/ViewControllers/MainViewController.cs#L59
and here 
https://github.com/toggl/mobile/blob/unidirectional/Ross/ViewControllers/MainViewController.cs#L219

the one you commented out was the one that closes the left panel from an open state.

You can experiment with this on emulator if you click drag on the header cell instead of TE cell.

Ross/ViewControllers/MainViewController.cs
```diff
@@ -21,7 +21,7 @@ public class MainViewController : UINavigationController
         private UIView fadeView;
 
         private UITapGestureRecognizer tapGesture;
-        private UIPanGestureRecognizer mainPanGesture, leftPanGestue;
+        private UIPanGestureRecognizer mainPanGesture, leftPanGesture;
```
taavirehemagi: still have a reference for mainPanGesture here

taavirehemagi:
Good for me! Let's merge

PR #1629, opened 2016-05-30
Fix #1593

### LINKED ISSUE #1593: iOS 7.1.0: Automatic edition of stop time causes an error
A user is complaining about a design choice when editing the start and stop times of an entry.

> If I quickly start and stop an event and then seek to adjust times (to do when I need to add an entry or break an old one into two) I used to adjust the start time and then input the time spent. Now it immediately moves end time so if I go too far back and then move forward in time I get an error that start time must be before end time and i have not found a way round this.

Case 290576

A possible solution could to be no to push the stop time further than the current time.

alfonsogarciacaro:
WIP: Please review.

To prevent user confusion, don’t check time correctness whenever start or stop entries are changed, only when the full time entry is about to be saved.

Notes:

- Because of the changes the UI for each platform is now responsible to show an error if necessary. This is because sometimes it doesn’t make sense to show an error when trying to save (e.g., when the `EditTimeEntryVM.Save` is called by Android’s `OnPause`).

- As the way to save changes in iOS is just to tap Back, this button has been overriden in iOS.

- Also to avoid confusion, I have commented out the code in `Phoebe.Helpers.DataExtensions.ChangeStartTime` that automatically modified the StopTime too. If the team agrees with the change, this code can be removed completely.


Ross/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -233,6 +233,22 @@ public override void ViewDidLoad()
...
+                {
+                    XPlatUtils.ServiceContainer.Resolve<GalaSoft.MvvmLight.Views.IDialogService>()
+                              .ShowMessage(EditTimeEntryVM.StartTimeError, EditTimeEntryVM.ErrorTitle);
+                }
+                else
+                {
+                    NavigationController?.PopViewController(true);
+                }
+            });
+            NavigationItem.LeftBarButtonItem = newBackButton;
```
monday8am: 👍

amulware:
I believe the webapp does not change the stop time anymore when changing the start time. On Desktop, this is a use setting. It is probably more consistent to not change it - or eventually make it a setting.

Regarding the OnPause-silent-save: Does this mean that if the user pauses the app, and then closes it (or it gets closed by the system for whatever reason), the time entry is saved in an invalid state?

taavirehemagi:
Works good for me. Maybe you can delete the commented-out lines though?

alfonsogarciacaro:
@amulware: the entry won't be saved with an invalid state. You can check the `EditTimeEntryVM.Save` method. Now the UI is responsible of showing the error, but the ViewModel still takes care the entry is not saved in an invalid state.

@taavirehemagi: yes, we can delete the commented out lines (I checked and `DataExtensions.ChangeStartTime` is not referenced by anything else) but if we want to use them in the future (for example, as @amulware says, if we want to make this a setting) maybe it's easier to leave them for the moment. I'm in favour of deleting them though to keep the code cleaner.

taavirehemagi:
We can always revert or look up code in git history. I don't see any problem with deleting it :)

taavirehemagi:
This is also good! let's merge this one as well

PR #1630, opened 2016-05-30
iPhone 4 compatibility

taavirehemagi:
The problem of projects screen not opening:
http://stackoverflow.com/questions/18746929/using-auto-layout-in-uitableview-for-dynamic-cell-layouts-variable-row-heights/18746930#18746930

The problem of workspace selector crashing the app is that UIPopoverPresentationController is not present in iOS 7, so check for a class first. If no class present, +1 the currentWorkspaceIndex #quickfix

### DIFF
--- a/Ross/ViewControllers/ProjectSelectionViewController.cs
+++ b/Ross/ViewControllers/ProjectSelectionViewController.cs
@@ -119,13 +119,13 @@ namespace Toggl.Ross.ViewControllers
 
                 if (data is ProjectData)
                 {
-                    var cell = (ProjectCell)tableView.DequeueReusableCell(ProjectCellId, indexPath);
+                    var cell = (ProjectCell)tableView.DequeueReusableCell(ProjectCellId);
                     cell.Bind((ProjectsCollectionVM.SuperProjectData)data, viewModel.ProjectList.AddTasks);
                     return cell;
                 }
                 else
                 {
-                    var cell = (TaskCell)tableView.DequeueReusableCell(TaskCellId, indexPath);
+                    var cell = (TaskCell)tableView.DequeueReusableCell(TaskCellId);
                     cell.Bind((TaskData)data);
                     return cell;
                 }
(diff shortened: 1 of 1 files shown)


Ross/ViewControllers/ProjectSelectionViewController.cs
```diff
@@ -119,13 +119,13 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
 
                 if (data is ProjectData)
                 {
-                    var cell = (ProjectCell)tableView.DequeueReusableCell(ProjectCellId, indexPath);
+                    var cell = (ProjectCell)tableView.DequeueReusableCell(ProjectCellId);
```
monday8am: why these changes? Add a comment or a link to some explanation.

Ross/ViewControllers/ProjectSelectionViewController.cs
```diff
@@ -97,8 +97,22 @@ private void OnAddNewProject(object sender, EventArgs evt)
...
+
+            bool hasPopover = ObjCRuntime.Class.GetHandle("UIPopoverPresentationController") != IntPtr.Zero;
+            if (hasPopover)
+            {
+                var popoverController = new WorkspaceSelectorPopover(viewModel, sourceRect);
+                PresentViewController(popoverController, true, null);
+            }
+            else
+            {
+                var nextWorkspace = viewModel.CurrentWorkspaceIndex + 1;
```
monday8am: seems to be you iterate over the list of workspaces. Add a comment please.

taavirehemagi: because we cannot show the picker for earlier iOS version, I had to come up with an alternative way to change between workspaces in iOS 7. it goes to the next workspace with every click now. (works the same in later versions)

Ross/ViewControllers/ProjectSelectionViewController.cs
taavirehemagi: I can't find the link at the moment but if I remember correctly, the function with the indexPath was introduced a bit later and might not be supported in earlier version.

PR #1634, opened 2016-05-31
Added TLP libraries and buffer to SyncManager.

monday8am:
Task Parallel Library
https://msdn.microsoft.com/es-es/library/hh228603(v=vs.110).aspx


Phoebe/Reactive/SyncManager.cs
```diff
@@ -96,12 +97,23 @@ public static void Cleanup()
             networkPresence = ServiceContainer.Resolve<INetworkPresence> ();
             client = ServiceContainer.Resolve<ITogglClient> ();
 
+            // TPL block to buffer messages and
```
alfonsogarciacaro: Is this a "To be continued" comment? ;)

Ross/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -177,11 +177,18 @@ public override void LoadView()
...
             View = scrollView;
 
             DescriptionEditingMode = false;
         }
 
+        private void backButtonPressed()
+        {
+            Console.WriteLine("here!");
+
+        }
```
alfonsogarciacaro: Interesting method :)

monday8am: upps!

Phoebe/Reactive/SyncManager.cs
```diff
@@ -96,12 +97,24 @@ public static void Cleanup()
...
+            }, blockOptions);
+
             StoreManager.Singleton
             .Observe()
 #if __TESTS__
             .SelectAsync(EnqueueOrSend).Subscribe();
 #else
-            .SubscribeQueued(EnqueueOrSend);
+            .Subscribe(processingBlock.AsObserver());
 #endif
```
alfonsogarciacaro: I run the tests directly with the action block and they're working so we may remove the __TESTS__ compiler directive (L114-116).

alfonsogarciacaro:
I tested it with the simulator (forcing continuation of entries very quickly) and seems to work. We must check it with a physical device.

PR #1637, opened 2016-06-02
More functional way to throttle autocomplete.

monday8am:


Phoebe/ViewModels/EditTimeEntryVM.cs
```diff
@@ -75,11 +72,13 @@ public EditTimeEntryVM(AppState appState, Guid timeEntryId)
...
-                      .Throttle(TimeSpan.FromMilliseconds(LoadSuggestionsThrottleMilliseconds))
-                      .ObserveOn(SynchronizationContext.Current)
-                      .Subscribe(async desc => await SuggestEntries(desc));
+            .Select(ev => ev.EventArgs)
+            .Where(desc => desc != null && desc.Length >= LoadSuggestionsCharLimit)
+            .Throttle(TimeSpan.FromMilliseconds(LoadSuggestionsThrottleMilliseconds))
+            .Select(desc => LoadSuggestions(desc))
+            .SubscribeOn(Scheduler.Default)
+            .ObserveOn(SynchronizationContext.Current)
+            .Subscribe(result => SuggestionsCollection = result);
```
alfonsogarciacaro: Shouldn't be the other way around? You're observing on the current context (which, as this code is run in the constructor, **should be** the UI thread) but you are subscribing in the `Scheduler.Default` context which (again) **should be** the task pool (i.e. background).

Moreover, `ObserveOn` comes after `.Select(desc => LoadSuggestions(desc))` which does the heavy work so it's not being moved in the background (well, in this case it is, but just because apparently `Throttle` is continued in the background).

All in all, the thread semantics of this workflow is not very clear and depends on a deep knowledge of how Rx works.

More info about how the SubscribeOn-ObserveOn combo works: http://www.introtorx.com/content/v1.0.10621.0/15_SchedulingAndThreading.html#SubscribeOnObserveOn

monday8am: > ObserveOn comes after .Select(desc => LoadSuggestions(desc)) which does the heavy work so it's not affecting it 
you're right. I was not clear with the order. I'll fix it. 👍 

> All in all, the thread semantics of this workflow is not very clear and depends on a deep knowledge of how Rx works.
I could replace the `and` by an `or`. If we use Rx, the idea is to use it extensively and some knowledge will be required for complex things like thread.

monday8am:
@alfonsogarciacaro we could merge this code if you don't have more comments..

alfonsogarciacaro:
Well, my comments are still the same as they were in the beginning (though they seem to have been deleted by a `push --force`). I'm not sure that code will work as intended, my proposal would be:

```
Observable.FromEventPattern<string>(h => DescriptionChanged += h, h => DescriptionChanged -= h)
.Select(ev => ev.EventArgs)
.Where(desc => desc != null && desc.Length >= LoadSuggestionsCharLimit)
.Throttle(TimeSpan.FromMilliseconds(LoadSuggestionsThrottleMilliseconds))
// Change to the task pool right before the time consuming op (LoadSuggestions)
.ObserveOn(TaskPoolScheduler.Default)
.Select(desc => LoadSuggestions(desc))
// Go back to current context (UI thread)
.SubscribeOn(SynchronizationContext.Current)
.Subscribe(result => SuggestionsCollection = result);
```

As this code is run in the ViewModel constructor, `SynchronizationContext.Current` **should be** the UI thread. I could be also set explicitly to make the intention cleared.

monday8am:
@alfonsogarciacaro I followed your suggestion to change the order.
https://github.com/toggl/mobile/blob/1a9407f3085e82b21b199811f7f0ee1f0988a662/Phoebe/ViewModels/EditTimeEntryVM.cs#L74

 I'll check `TaskPoolScheduler.Default`, seems to the `Scheduler.Default` pick the most suitable scheduler but really I haven't discovered the TaskPoolScheduler, I'll replace it.

In your example code `ObserveOn` and `SubscribeOn` are used in a reversed way I think.

alfonsogarciacaro:
I've checked and I my understanding about how `SubscribeOn` and `ObserveOn` work was wrong (again). If [this information](http://stackoverflow.com/a/37635953) is right, `SubscribeOn` affects the whole pipeline (I thought it only affected the lambda in `Subscribe`) while `ObserveOn` only affects the steps below it. That's why `SubscribeOn` usually appears on top.

```csharp
Observable.FromEventPattern<string>(h => DescriptionChanged += h, h => DescriptionChanged -= h)
// Observe on the task pool to prevent locking UI
.SubscribeOn(TaskPoolScheduler.Default)
.Select(ev => ev.EventArgs)
.Where(desc => desc != null && desc.Length >= LoadSuggestionsCharLimit)
.Throttle(TimeSpan.FromMilliseconds(LoadSuggestionsThrottleMilliseconds))
.Select(desc => LoadSuggestions(desc))
// Go back to current context (UI thread)
.ObserveOn(SynchronizationContext.Current)
.Subscribe(result => SuggestionsCollection = result);
```

About `TaskPoolScheduler.Default`, again is a matter of semantics. We don't know exactly what is the `Scheduler.Default` but it's more evident that `TaskPoolScheduler` will put the work in the background... unless I'm completely mistaken by the names as it usually happens with Rx :)

monday8am:
already added at #1691

PR #1644, opened 2016-06-06
New iOS timer screen design

amulware:


Ross/Info.plist
```diff
@@ -56,6 +56,8 @@
 	<key>UIAppFonts</key>
 	<array>
 		<string>Fonts/icomoon.ttf</string>
+		<string>Fonts/SF-UI-Text-Light.otf</string>
+		<string>Fonts/SF-UI-Text-Regular.otf</string>
```
monday8am: I think we can't use San Francisco font in that way. Maybe is better to use the *system font* utility.
`v.Font = UIFont.SystemFontOfSize(14f, UIFontWeight.UltraLight);`

http://stackoverflow.com/questions/33096954/san-francisco-font-in-ios-8

amulware: However, this means that for older versions of iOS the app will look different, right? Are we ok with this?

monday8am: they will use their default system font. I'm afraid if we include the SF font without the proper license, Apple rejects the binary. But we can take the risk if you want.

amulware:
@monday8am consider this a resubmit. As far as I'm concerned we can merge this.

monday8am:
thanks @amulware, I'll review it ASAP 👍

amulware:
@monday8am bump :P

monday8am:
except for the use of ".this" :P , for me is ok. Good job!

PR #1645, opened 2016-06-06
Merge previous no-user-code with existing Unidirectional branch

monday8am:


monday8am:
canceled. rebased was needed. replaced by #1652

PR #1649, opened 2016-06-07
Solution to show splash logo correctly.

monday8am:
@taavirehemagi another solution, how you tested previously? (which device failed?)


taavirehemagi:
This solution works 👍 
Previously, on some devices the left and right edges of the screen were black for ~10px on either side to keep the pictures aspect ratio. ( a normal emulator for example 768x1280) 

anyway. This is ready to merge from my part.

PR #1650, opened 2016-06-07
Hide welcome view before show Timer screen.

monday8am:
@taavirehemagi please, let me know if it fits the bug you described.


taavirehemagi:
What I meant was, that even though there are time entries, the empty state overlay doesn't disappear. 
Start app (fresh state) -> Login ->
screen shot 2016-06-08 at 10 03 19

Also, when I add a time entry with manual mode, the overlay doesn't disappear. Only way it does disappear is when I click the start button

monday8am:
mm... interesting, let me check..

monday8am:
@taavirehemagi please, check it the new commit to see if it works.

taavirehemagi:
The problem still occurs the same way. 
Steps to reproduce:
1. Uninstall app.
2. Reinstall app (clean state)
3. Sign in (for example with test.account.2@toggl.com)
4. Directs to Timer page. Time entries start loading but still the Overlay is also present.
screen shot 2016-06-09 at 15 40 43

taavirehemagi:
@monday8am this is still open and needs to be fixed and merged to nouser branch

taavirehemagi:
@pe0ny @kyllikekimmel  Also, I think this shouldn't be the empty state screen when the user is logged in, since it says "Log in" at the top

monday8am:
Changed target branch.

PR #1652, opened 2016-06-07
No-user-mode branch rebased and with migration content.

monday8am:


taavirehemagi:
This doesn't have the latest two commits from the unidirectional-nouser branch (which also is rebased with unidirectional branch. I can make a PR about nouser and then you can make a PR about migration, I think that's a bit simpler?

monday8am:
replaced by #1714

Issue #1538, opened 2016-05-10
Allow editing duration of running time entries


amulware:
In web and desktop, this can be done, and the app adjusts the start-time of the entry appropriately.

Makes sense to me to allow this in the mobile app as well.

Personally I like being able to say "I've been working on this for 15 minutes" without having to do that math in my head.

monday8am:
@amulware use tags :P

PR #1661, opened 2016-06-09
improved ios duration and manual-entry editing, fixes #1632, #1538

### LINKED ISSUE #1632: Manual entry screen creates an entry with value 0.
Steps:
- Press Add Manual btn (the top number)
- Press Back.
- A new entry with duration 00.00.00 is added. 

with Back we should not create anything, IMO.

amulware:
Does the following:
- manual entries now only saved if user enters any details (duration defaults to 0!)
- allows time entries to have start time == stop time
- allows editing of duration of running time entry
- made editing duration of running time entry, or manual time entry keep stop-time fixed and change start time instead


amulware:
This should be against `ios7`

PR #1662, opened 2016-06-09
Fix #1660

### LINKED ISSUE #1660: SQLite.Net.SQLiteException Busy. Failed to handle DataMsg: XXXXXX
Description:
https://insights.xamarin.com/app/Toggl-iOS/issues/1584?s=SQLiteException

The bug occurs generally for messages like:
`TimeEntryStart`, `TimeEntryStop`, `ServerResponse` because are the most common operations.

Maybe a buffer is needed.

alfonsogarciacaro:
Attempt to fix #1660 by doing the following:

- Add a buffer to `StoreManager` as with `SyncManager` to be sure all events are processed sequentially.
- Use a different file for the queue, so `StoreManager` and `SyncManager` don’t try to write the db at the same time.

### DIFF
--- a/Phoebe/Data/DatabaseHelper.cs
+++ b/Phoebe/Data/DatabaseHelper.cs
@@ -10,6 +10,8 @@ namespace Toggl.Phoebe.Data
 {
     static class DatabaseHelper
     {
+        public const string QueueExtension = ".queue.db";
+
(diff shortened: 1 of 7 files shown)


monday8am:
ok for me 👍

PR #1667, opened 2016-06-09
Removed single requirement, removed id from warnings.

monday8am:


taavirehemagi:
Why did you remove the ID from the logging?

monday8am:
because it generates 100xxxx warnings of different types and I can't do anything with a local ID.

taavirehemagi:
makes sense 👍

PR #1691, opened 2016-06-20
Improvements from iOS 7 milestone

monday8am:


Phoebe/Reactive/SyncManager.cs
```diff
@@ -193,13 +193,11 @@ void logWarning(string msg, Exception exc = null)
                         }
                         catch (Exception ex)
                         {
-                            if (ex is RemoteIdException)
-                                logInfo(ex.Message);
-                            else
-                                logError(ex);
-
+                            // TODO After a bug, be optimistic and expect that
+                            // everything will be ok again?
```
monday8am: @amulware I remember your comment about this point. Alfonso fixed this part, I'll review it again.

Ross/ViewControllers/ProjectSelectionViewController.cs
```diff
@@ -24,11 +24,11 @@ public class ProjectSelectionViewController : UITableViewController
...
         private readonly IOnProjectSelectedHandler handler;
 
-        public ProjectSelectionViewController(Guid workspaceId, IOnProjectSelectedHandler handler) : base(UITableViewStyle.Plain)
+        public ProjectSelectionViewController(EditTimeEntryViewController editView) : base(UITableViewStyle.Plain)
         {
             Title = "ProjectTitle".Tr();
-            this.workspaceId = workspaceId;
-            this.handler = handler;
+            this.workspaceId = editView.WorkspaceId;
+            this.handler = editView;
```
monday8am: @amulware why use a ViewController here as parameter and not a `Guid` and an interface?

amulware: No particular reason as far as I can see. Probably an artifact of earlier changes, or changes that were not committed for other reasons. Feel free to change this if you prefer.

PR #1693, opened 2016-06-20
Feature/remove duplicated activeentry

xleon:
This PR is trying to fix a bug where the user is getting more than one active time entry at the same time. Given the error is not reproducible I think this can happen under circumstrances where the dictionary is accessed/modified at the same time from different places. I converted the TimeEntries to a ConcurrentDictionary and also implemented a lock in the ActiveEntry getter. I´m throwing an exception instead of just logging to get a crash log if this happens again


xleon:
Got it. I´ll see what I can do

Issue #1653, opened 2016-06-08
Create new loading splash screen and animation


amulware:
Splash screen will be solid white/light gray with the toggl logo and text on it (slightly larger than the one on the timer page).

![toggl ios - splash 1x](https://cloud.githubusercontent.com/assets/3987892/15884454/e5c640a8-2d50-11e6-9d81-79495f1f4d9a.png)

When loading finishes, the background will fade away, while the logo will move and scale (smoothly, i.e. animated) to the top of the timer list.

Create on new branch, branching off of `paul-ios-new-timer-screen`. This branch already includes the new design which has the logo at the top of the timer list, serving as end point for the animation

pe0ny:
Assets are available in Zeplin ;)

PR #1694, opened 2016-06-21
Feature/splash animated

xleon:
animation


Ross/ViewControllers/MainViewController.cs
```diff
@@ -133,29 +129,48 @@ private bool tryMigrateDatabase(IUserData userData)
...
+                    UIView.Animate(0.28, 0.3, UIViewAnimationOptions.CurveLinear,
+                                   () => { splashLogo.Alpha = 0; splashBackground.Alpha = 0; },
+                                   () => { splashLogo.RemoveFromSuperview(); splashBackground.RemoveFromSuperview(); });
+                 });
             }
             else
             {
-                vc = new WelcomeViewController();
                 MenuEnabled = false;
+                splashLogo.RemoveFromSuperview();
```
amulware: The splashBackground also has to be removed in the logged out case, otherwise it causes the app to be unusable. ;)

xleon: done

amulware:
Overall looking pretty good!

Just one question:

Currently the logo animates to the top, and afterwards the background fades away to reveal the timer. From a user's point of view this feels sort of like a two-step loading thing to me, which is a bit weird.

Is there some reason that we cannot have both animations at the same time, @xleon?

xleon:
I was doing it that way as the first option. But this [line](https://github.com/toggl/mobile/pull/1694/files#diff-5c5d66cb6410fc07971d4e05fdeb1bddR155) is blocking the ui thread then the animation suffers becoming not smooth at all. So I decided to run that code after the logo animation is finished.

**Possible solution**: run that line of code before animation starts (not sure if that will have a negative effect somewhere else though)

xleon:
https://github.com/toggl/mobile/issues/1653

PR #1695, opened 2016-06-21
Bugfix/remove duplicated active entries

xleon:


Phoebe/Reactive/Reducers.cs
```diff
@@ -432,10 +434,12 @@ static DataSyncMsg<AppState> TimeEntryContinue(AppState state, DataMsg msg)
             var updated = dataStore.Update(ctx =>
             {
                 // Stop ActiveEntry if necessary
-                var prev = state.ActiveEntry.Data;
-                if (prev.Id != Guid.Empty && prev.State == TimeEntryState.Running)
+                var activeEntry = state.TimeEntries.FindActiveEntry();
+                if(activeEntry != null 
+                   && activeEntry.Data.Id != Guid.Empty 
+                   && activeEntry.Data.State == TimeEntryState.Running)
```
monday8am: @xleon I think we can move this condition:
`activeEntry.Data.Id != Guid.Empty`
to the `FindActiveEntry` extension and only check that the result is null or not.

Phoebe/ViewModels/LogTimeEntriesVM.cs
```diff
@@ -62,59 +66,54 @@ public LogTimeEntriesVM(AppState appState)
...
+                .Subscribe(x => UpdateSettingsAndRequestInfo(x.Settings, x.RequestInfo));
+
+            subscriptionTimeEntries = StoreManager
+                .Singleton
+                .Observe(x => x.State)
+                .ObserveOn(uiContext)
+                .StartWith(appState)
+                .Select(state => state.TimeEntries.Values.Where(e => e.Data.State == TimeEntryState.Running))
+                .DistinctUntilChanged(x => x.Count())
+                .SubscribeOn(TaskPoolScheduler.Default)
```
monday8am: in theory, the `SubscribeOn` is not needed because every update in the state is done in a background context.

Phoebe/ViewModels/LogTimeEntriesVM.cs
```diff
@@ -170,27 +171,24 @@ public void RemoveTimeEntry(int index)
...
-        }
+        public bool IsInExperiment() => 
+            OBMExperimentManager.IncludedInExperiment(StoreManager.Singleton.AppState.User);
 
-        public bool ShowWelcomeScreen()
-        {
-            return StoreManager.Singleton.AppState.Settings.ShowWelcome;
-        }
+        public bool WelcomeScreenShouldBeShown => 
+            StoreManager.Singleton.AppState.Settings.ShowWelcome;
```
monday8am: good point! 👍

Phoebe/ViewModels/LogTimeEntriesVM.cs
xleon: As I understand, `SubscribeOn` will run `UpdateSettingsAndRequestInfo` in another thread. Is that correct? This controller is blocking the ui thread on initialization (I found this out when animating the logo)

Phoebe/Reactive/Reducers.cs
xleon: I didn´t see that one :)

PR #1696, opened 2016-06-22
Bugfix/diego active entry

xleon:


monday8am:
we'll need  a new PR with this content. I'll create a version for the tester user.

PR #1698, opened 2016-06-22
Bugfix/billable by default

xleon:


Phoebe/Data/SyncSqliteDataStore.cs
```diff
@@ -89,7 +90,16 @@ private void CleanOldDraftEntry()
...
+        }
+
+        // temporal hack to avoid migrations for adding simple columns/fields to models
+        private void EnsureNewColumnsExists()
+        {
+            // var map = cnn.GetMapping<WorkspaceData>().Columns.Select(x => new { x.Name, x.PropertyName });
+            // var workspaces = cnn.Table<WorkspaceData>();
+            // Make sure WorkspaceData.ProjectsBillableByDefault column is added to the db
+            cnn.CreateTable<WorkspaceData>();
+            // TODO retrieve all workspaces from server and update local db
```
xleon: @monday8am This is what I need to do

Phoebe/Net/ITogglClient.cs
```diff
@@ -44,6 +43,10 @@ Task Delete<T> (string authToken, IEnumerable<T> jsonObjects)
 
         Task<List<WorkspaceUserJson>> ListWorkspaceUsers(string authToken, long workspaceId);
 
+        Task<List<WorkspaceJson>> ListWorkspaces(string authToken);
+
+        Task<WorkspaceJson> GetWorkspace(string authToken, long id);
```
monday8am: you can use a generic method `Get<WorkspaceJson> (authToken, id) . A new method isn't needed.

xleon: The method was already done in the implementation. But you are right. Even the implementation should be removed

Phoebe/Reactive/SyncManager.cs
```diff
@@ -148,6 +148,12 @@ void logWarning(string msg, Exception exc = null)
 
         async Task HandleRequest(ServerRequest request, AppState state)
         {
+            if(request is ServerRequest.GetWorkspaces)
+            {
+                await GetWorkspaces(request, state);
+                return;
+            }
+
```
monday8am: is better to use here the `MatchType` like with the other messages IMO. It is already implemented.

xleon: True. I´m doing some spagetti code to have this working. Then I will do it properly.

xleon:
I tested this in iOS emulator and it seams to be working like a champ. 
Needs some testing in devices + android

PR #1700, opened 2016-06-23
changed continue swipe button font to be in line with system delete button

amulware:


amulware:
Styling the delete button is... annoying and hacky. Luckily it basically looks like we want, so we can just make the continue/stop swipe buttons look the same (in this case just a matter of font height) and everything is mostly fine.

PR #1741, opened 2016-07-18
update experiment m83, closes #1710

### LINKED ISSUE #1710: Start iOS OBM experiment M83 from scratch
Since iOS design changed we'll start OBM experiment #1062 from scratch.
Design and assets are available in Zeplin.

* Go live when #1282 is ready
* Display text “Welcome! Start your timer here” with an arrow pointing towards the green "Start" icon on the timer page until the user has created his/her first time entry.
* If user has seen the experiment, created time entries and deletes all time entries, don’t show the experiment anymore.
* If user switches to manual mode, keep the experiment visible. 
* Display it to all users who have signed up from mobile and who are in "experiment_group = included" group.
* When the user starts his/her first time entry from the green ​*”Start"*​ button, track it as ​*experiment_actions = “startbutton”*​ (add those actions to both included and excluded groups to compare afterwards)
* When the user starts his/her first time entry from the ​*”Add"*​ button in manual mode, track it as ​*experiment_actions = “addbutton”*​ (add those actions to both included and excluded groups to compare afterwards)

Melissa or Küllike will test the experiment before we go live!

toggl ios - obm ver2

amulware:


Phoebe/ViewModels/EditTimeEntryVM.cs
```diff
@@ -28,6 +28,8 @@ public class EditTimeEntryVM : ViewModelBase, IDisposable
         private RichTimeEntry previousData;
         private event EventHandler<string> DescriptionChanged;
 
+        public Action OnSaveManuel { get; set; } // for experiments
```
monday8am: typo here, except for that, everything looks ok for me. You can let it unchanged if you want to trick spanish speakers :)

https://en.wikipedia.org/wiki/Manuel_(name)

amulware: Thanks! :D

PR #1769, opened 2016-07-27
Adds Top Projects header

### LINKED ISSUE #1764: Show top projects in project selection view
Already implemented in shared code and Android.

willsb:
Closes #1764


amulware:
There currently is no margin between the last top project and the first client-section-header ('No Client'). Let's add some margin there, and in return decrease the margin around the 'top projects' head.
screen shot 2016-07-28 at 11 27 09

Ross/ViewControllers/ProjectSelectionViewController.cs
```diff
@@ -62,7 +66,49 @@ public override void ViewDidLoad()
...
             }
+
             TableView.TableFooterView = null;
+
+            //Adds the top projects header
+
+            //Enumerates only once
+            var topProjects = viewModel.TopProjects.ToList();
+
+            var numberOfProjects = topProjects.Count();
```
amulware: Nitpick: this can use the property `Count` instead of the LINQ method.

amulware:
Ah, and I found a much more serious problem: The top projects update when switching workspaces (they do not right now). Compare with android, where they do.

Also the top projects should ideally display the client and task names, similar to android.

Ross/ViewControllers/ProjectSelectionViewController.cs
```diff
@@ -69,25 +68,37 @@ public override void ViewDidLoad()
...
-            var topProjects = viewModel.TopProjects.ToList();
+            var topProjects = viewModel.TopProjects?.ToList();
 
-            var numberOfProjects = topProjects.Count();
-            if (numberOfProjects == 0) return;
+            var numberOfProjects = topProjects?.Count ?? 0;
+            if (numberOfProjects == 0)
+            {
+                if (TableView.TableHeaderView == null) return;
+                UIView.BeginAnimations(null);
```
amulware: What does this do here?

amulware:
One detail: With the two lines of text, [descenders](https://en.wikipedia.org/wiki/Descender) are cut off.
screen shot 2016-07-28 at 14 45 24

Ross/ViewControllers/ProjectSelectionViewController.cs
willsb: Null coalescing. Just in case.

PR #1777, opened 2016-07-28
No user mode in iOS

willsb:


monday8am:
nice job! 👍

Ross/ViewControllers/LogViewController.cs
```diff
@@ -95,11 +95,49 @@ public override void LoadView()
                 Message = "LogEmptyMessage".Tr(),
             };
 
-            obmEmptyView = new OBMEmptyView
-            {
-                Title = "LogOBMEmptyTitle".Tr(),
-                Message = "LogOBMEmptyMessage".Tr(),
-            };
+            noUserEmptyView = new UIView();
```
amulware: This view should probably be abstracted into a separate class.

amulware:
Apart from the comment there are a bunch of other things I noticed that I'll bring up later, but overall, this is looking pretty good. Carry on!

willsb:
Current state of this PR:

1 - The user is taken straight to the empty state with the little toggler dude, where he can start tracking time without signin-in.
2 - The login flow is working fine.
3 - The empty states for screen that should be disabled in No user mode are currently using mock layouts, but their functionality is good to go.
4 - The drawer states are being handled accord to whether the user is authenticated or not.

Issues:
1 - When the user tries to sign up, a null reference exception is thrown by [Phoebe/Reactive/Reducers.cs](https://github.com/toggl/mobile/blob/baf2e42669609a23fa27d68cc8443ec4c029346d/Phoebe/Reactive/Reducers.cs#L453), on line 453. This happens because there are no workspaces in the AppState. This is also noted when one tries to create a new Project before signing in: the app crashes when you try to do so. This is probably the missing bit of information that is missing.
2 - It seems that the TEs for newly created accounts are not being synced to the server when the user logs out. I'm not sure if this is something from the development server or not, so I think it's worth checking it as well.

amulware:
The above issues still have to be addressed, and extensive testing has to be done as well.
Further, designs for the empty reports and feedback views are in the pipe and should arrive shortly.

Merging this for now to have it in the main repo.

Issue #1805, opened 2016-08-10
iOS top projects


pe0ny:
iOS top projects design should be changed accordingly:

![toggl ios - projects 2x](https://cloud.githubusercontent.com/assets/5064445/17552699/04bab2c4-5f0a-11e6-9561-1de513144fe2.png)

- Top three projects should look the same like other projects list
- Top projects should be updated when switching between workspaces (top projects are workspace based)
- Difference between top projects and other projects is that we have to show client name (if any)
- This functionality change won't change android functionality

Related issues #1769 

Check zeplin for additional details

amulware:
Apart from the styling, the main difference to the current implementation is that currently we do not have collapsing tasks, while with this new design, we do.

thefex:
As discussed on Slack - expand/collapse of Top Projects is not currently implemented.
Pull Request contains (mentioned above):
- new UI (top projects look like normal project rows)
- search functionallity ( lacks auto-hide of SearchBar when user is "searching and scrolling" )

willsb:
@pe0ny Does the collapsible tasks part make sense? Data for those projects already includes the most used task, so if the user has two different tasks in top projects, the same project will appear twice.

amulware:
This issue has not been updated according to internal discussions: We decided that having collapsing projects would take too much effort.

pe0ny:
@willsb you can check zeplin for latest designs.

Also, top project can't contain more than one task. Which means user can add more than one task to a project but we'll display only most popular ones. And this means we can display three times the same project but with different tasks.

And the order will be:
Project - Client
Task

PR #1809, opened 2016-08-11
Issue #1805

thefex:
Implemented top Projects header UI according to mockups. 
Implemented search functionallity - iOS.

### DIFF
--- a/Ross/Ross.csproj
+++ b/Ross/Ross.csproj
@@ -20,7 +20,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <ConsolePause>false</ConsolePause>
-    <CodesignEntitlements>Entitlements.plist</CodesignEntitlements>
+    <CodesignEntitlements></CodesignEntitlements>
     <MtouchDebug>true</MtouchDebug>
     <MtouchI18n>
     </MtouchI18n>
--- a/Ross/Theme/Style.ProjectList.cs
+++ b/Ross/Theme/Style.ProjectList.cs
@@ -22,35 +22,35 @@ namespace Toggl.Ross.Theme
 
             public static void ProjectLabel(UILabel v)
             {
-                v.Font = UIFont.FromName("HelveticaNeue", 17f);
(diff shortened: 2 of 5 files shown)


amulware:
This still needs the following changes:

- Top projects should disappear while showing a filtered list
- While searching, the search bar should be floating at the top, even if the user scrolls through the list

Ross/Ross.csproj
```diff
@@ -20,7 +20,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <ConsolePause>false</ConsolePause>
-    <CodesignEntitlements>Entitlements.plist</CodesignEntitlements>
+    <CodesignEntitlements></CodesignEntitlements>
```
amulware: Please remove this change.

Ross/Theme/Style.ProjectList.cs
```diff
@@ -22,35 +22,35 @@ public static void HeaderLabel(UILabel v)
 
             public static void ProjectLabel(UILabel v)
             {
-                v.Font = UIFont.FromName("HelveticaNeue", 17f);
+                v.Font = UIFont.FromName("HelveticaNeue", 15f);
```
amulware: Please use the `Font` class to obtain the correct font (and same below).

Ross/Ross.csproj
thefex: This just remove need for provisioning profile on iPhone simulator. The change is done only for "simulator" debug profile.

thefex:
"Top projects should disappear while showing a filtered list" - this is related to other issue, fixed by one of my pull requests for Android. After merging Android PR with fix - it should be quick fix with few lines of iOS code.

Ross/Ross.csproj
amulware: Fair enough.

amulware:
The referenced pull request was merged, and top projects disappear on android when searching as desired. As @thefex says, making this work in iOS as well should not be hard.

@yaremam Please merge/rebase the changes here locally and implement that behaviour (and other comments) for a new PR. :)

PR #1810, opened 2016-08-15
Ios no user

yaremam:
- can add projects/clients when no user is logged in
- signup works properly
- in case of invalid login attempt default workspace/user created
- after the user logs out switch to default user/workspace


amulware:
Please make PR against the `ios-no-user` branch.

PR #1811, opened 2016-08-15
No user iOS improvements

yaremam:
- can add projects/clients when no user is logged in
- signup works properly
- in case of invalid login attempt default workspace/user created
- after the user logs out switch to default user/workspace
-TEs created before signing up will be pushed to the server when the user logs out


amulware:
Did some testing, overall it's looking pretty good, but there are some issues with logging in/signing up. Specifically, it seems that multiple logins/outs/signups/outs causes that part to break somehow, so I cannot sign up anymore - though instead of getting an error, the sign up screen just freezes. In addition, if I try logging in with wrong password, it goes straight back to the empty view, without any error message.

Further, I'd like to double check the styling and layout of the no-user feedback and reports screens.

amulware:
Merging this for now to get the improvements into our repo. More work has to be done to address the above issues before this can be merged into `unidirectional`.

yaremam:
Logging in with wrong password was not displaying error message before the changes either, but only for the first time. Second time you try to login with wrong password the error message was shown. Will check it again later. Might be related to the fact that Reset is calling NoUserDataPut. Alternative would be to send both Reset and NoUserDataPut after logout, but this causes unnecessary reload of the UI.

willsb:
I couldn't reproduce the signup issue, @amulware. I'll merge this for now to continue the fixes

